PR #21, opened 2023-11-28
feat: dockerfile

y-eight:
## Motivation

Prepared Dockerfile for building the `sparrow` image.

## Changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
-ARG GO_VERSION
+FROM golang:1.21-bullseye as gobuild
 ARG VERSION
-
-FROM golang:${GO_VERSION}-bullseye as gobuild
 RUN adduser \
     --disabled-password \
     --shell "/sbin/nologin" \
@@ -13,7 +11,7 @@ WORKDIR /app
 ADD . .
 RUN go mod download
 RUN go mod verify
-RUN CGO_ENABLED=0 go build -ldflags "-s -w -X main.version=${VERSION}" -o sparrow .
+RUN CGO_ENABLED=0 go build -ldflags '-s -w -extldflags "-static" -X main.version=${VERSION}' -o sparrow .
 
 FROM scratch
 COPY --from=gobuild /etc/passwd /etc/passwd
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM cgr.dev/chainguard/go:latest as gobuild
```
lvlcn-t, line 1: As @puffitos mentioned in https://github.com/caas-team/sparrow/pull/22#discussion_r1408883502

> Maybe use the go1.21 image equivalent, no need for latest :)

We should define the go version here

y-eight: since chainguards images by tag are not available for free I have switched to golang image for the app build.

y-eight: golang version will be passed to the dockerfile by --build-arg

NiklasTreml:
Scratch images come without certificates, please add them to the runner image, otherwise we won't be able to use https:
```dockerfile
FROM alpine:latest as certs
RUN apk add --no-cache ca-certificates
...
# In your runner image
COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
```

y-eight:
Ive added the ca certs directly from gobuild stage

puffitos (approved):
I dislike having a variable go version (we don't have to abstract/ leave everything to the pipeline), but other than that, lgtm.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid 65532 \
+    sparrow
+
+WORKDIR /app
+ADD . .
+RUN go mod download
+RUN go mod verify
+RUN CGO_ENABLED=0 go build -ldflags "-s -w -X main.version=${VERSION}" -o sparrow .
```
NiklasTreml, line 16: Please also add the `static` flag to the linker

```bash
CGO_ENABLED=0 go build -a -ldflags '-s -w -extldflags "-static" -X main.version=0.0.1' -o app ./main.go
```

NiklasTreml (changes requested):

PR #29, opened 2023-12-05
fix: latency test race condition

### LINKED ISSUE #27: Bug: Test race condition

### Current Behavior

Test for latency check is failing with race conditions.

See: https://github.com/caas-team/sparrow/actions/runs/7058951310/job/19215535279

Hints by @puffitos see: https://github.com/caas-team/sparrow/pull/26#issuecomment-1837234461

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

_No response_

### Anything else?

_No response_

lvlcn-t:
## Motivation

Address a race condition found within the `check` method of the Latency Check. The race condition was caused by the concurrent goroutines capturing the loop variable `e` from the `for` loop, leading to unpredictable behavior as this variable was shared across all goroutines.

Closes #27 

## Changes

Introduced a local copy of the loop variable in the `for` loop to ensure each goroutine receives a unique, stable copy of the target URL.

## Tests done

The unit tests succeed now.

## TODO


y-eight:
Race cond. still present: https://github.com/caas-team/sparrow/actions/runs/7099127501/job/19323035475

lvlcn-t:
Ready for review.

This PR now also partly addresses #31 but needs to be extended in another PR for the health check and other refinements.

pkg/checks/health.go
```diff
@@ -116,6 +116,11 @@ func (h *Health) SetConfig(_ context.Context, config any) error {
 	return nil
 }
 
+// SetClient sets the http client for the health check
+func (h *Health) SetClient(_ *http.Client) {
```
puffitos, line 120: @lvlcn-t I've removed the parameter here

pkg/sparrow/run_test.go
```diff
@@ -189,19 +190,29 @@ func Test_fanInResults(t *testing.T) {
...
 	}
 
-	s := New(c)
+	c.SetLoaderFilePath("../config/testdata/config.yaml")
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
-	defer cancel()
-	if err := s.Run(ctx); err != nil && !errors.Is(err, context.DeadlineExceeded) {
-		t.Errorf("Sparrow.Run() error = %v", err)
-	}
```
puffitos, line 206: No idea why this test was passing, it almost always failed in my IDE. I've made the the test a positive case; the API starts and no error is thrown.

puffitos (commented):
LGTM; @y-eight  / @NiklasTreml  I'll wait for your review as well before merging.

pkg/checks/latency.go
```diff
@@ -57,11 +59,8 @@ func (l *Latency) Run(ctx context.Context) error {
...
 		case <-time.After(l.cfg.Interval):
-			results, err := l.check(ctx)
+			results := l.check(ctx)
 			errval := ""
-			if err != nil {
-				errval = err.Error()
-			}
 			checkResult := Result{
 				Data:      results,
 				Err:       errval,
```
lvlcn-t, line 66: We could now remove this field as it is part of the `Data` field in both checks.

puffitos: Possibly, but not as part of this PR. Why is the error in the Data as well, if there's already an `Err` field on the general `Result` struct? Is it important to store extra information on the error in the `Data` itself? Why is the `Result.Err` a `string` and not an `error`? Let's solve those questions first and then we can definetely move on from there.

y-eight: Result.Err is a string because it is directly exposed by the Api.

Currently it is not used. We should consider using it when something in general is not working for the check (like in the registration part). We should not forget about this but should be part of [issue]().

y-eight (approved):
LGTM

PR #24, opened 2023-11-29
chore/linting

puffitos:
## Motivation

Close #23 

## Changes

Added linting with golangci, turned off some of the most strict linters. Also implemented most of the linter's suggestions. We should discuss some of the TODOs before merging (a lot of errors were left unchecked...)

## TODO

- [x] I've assigned this PR to myself
- [x] I've labeled this PR correctly

PR #39, opened 2023-12-08
docs: rm auto gen tag

lvlcn-t:
## Motivation

I deactivated the automatic generation tag at the bottom of the flag doc generation because it includes the current date and that means that the pre-commit hook (introduced with #24) fails once every day. Additionally, it produced overhead on the commits.

## Changes

I modified the `runGenDocs` function in the `cmd` package to disable the automatic generation tag at the bottom of the flag documentation. Specifically, I set `rootCmd.DisableAutoGenTag` to `true` in the `runGenDocs` function to prevent the inclusion of the current date in the documentation.

For additional information, please refer to the commits.

## Tests done

Everything generated just as intended.

## TODO


puffitos (approved):
Thanks for this change.

PR #40, opened 2023-12-08
fix: latency req timeout calc

lvlcn-t:
## Motivation

The latency check currently uses the wrong time format (µs instead of seconds).

## Changes

I've changed it to seconds and additionally pulled the timeout assignment out of the go routine since it's unnecessary to assign it every time.

## Tests done

Also added another test case for the run unit tests to test multiple targets.

All unit tests still succeed.

## TODO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ gen
 
 # test configs for debug launch configuration
 .vscode/config
+
+# Temporary directory
+.tmp/*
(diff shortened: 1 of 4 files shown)


NiklasTreml:
LGTM, but I think we should figure out a better system for this in the future, so we only do this when we parse the config and not inside the checks.

y-eight (approved):

PR #42, opened 2023-12-14
Fix/health shutdown

y-eight:
## Motivation

The health check shutdown did not work properly. Due to an uninitialized channel the shutdown could not be performed, the sparrow would have been blocked.

Additionally, I have rm the moq files from gitignore @puffitos. Should be pushed to repo.

## Changes

## Tests done

- unit test


puffitos (approved):

lvlcn-t (approved):
LGTM

lvlcn-t:
> LGTM, thanks for including the mocks.

Since we've now included the mocks, we can remove the generation from the [unit test job](https://github.com/caas-team/sparrow/blob/70243c35a8ccecdebe2b8b29995fe5a46952173b/.github/workflows/test_unit.yml#L26)

PR #54, opened 2024-01-03
feat(helm): add target manager support

y-eight:
## Motivation

With `v0.2.0` we have introduced the target manager. Therefore we needed to upgrade the helm chart. Here we go... 

## Changes

## Tests done

- deployed in k8s env
- E2E will fail. It needs to be refactored. Option to diable the target manager will be done with https://github.com/caas-team/sparrow/issues/55

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -36,6 +36,8 @@ spec:
           {{- if .Values.startupConfig}}
           - --config
           - /startupconfig/.sparrow.yaml
+          - --tmconfig 
+          - /startupconfig/tmconfig.yaml
           {{- else if .Values.extraArgs }}
           {{- range $key, $value := .Values.extraArgs }}
           - --{{ $key }}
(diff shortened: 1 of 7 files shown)


chart/templates/deployment.yaml
```diff
@@ -36,6 +36,8 @@ spec:
           {{- if .Values.startupConfig}}
           - --config
           - /startupconfig/.sparrow.yaml
+          - --tmconfig 
+          - /startupconfig/tmconfig.yaml
```
NiklasTreml, line 40: I think we should refactor this in the future, so we don't have a separate config file for the target manager

NiklasTreml (approved):

puffitos (approved):
lgtm

lvlcn-t (approved):
LGTM
---
I've additionally relinked the dead link in the [runtime config docs](https://github.com/caas-team/sparrow/tree/main?tab=readme-ov-file#runtime)

chart/templates/deployment.yaml
y-eight: Yes, this needs to be involved in the config refactoring

PR #57, opened 2024-01-04
feat: always activate health endpoint; rm unnecessary config

y-eight:
## Motivation

The health check endpoint has been configurable. There is no need to enable or disable it. Therefore the config option has been removed. The endpoint is always enabled as soon as the health check is active.

Additionally, the `enabled` config option of the health and latency check was not used at all. Therefor the option has been removed.

## Changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ apiVersion: 0.0.1
 kind: Config
 checks:
   health:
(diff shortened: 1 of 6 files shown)


NiklasTreml (approved):
LGTM

PR #58, opened 2024-01-04
Make target manager optional

### LINKED ISSUE #55: Option to Disable the Target Manager

### Problem Description

Currently, the Target Manager is enabled by default. The sparrow is working without the Target Manager fine. If the use case is not needed it is not possible to disable the TM.

### Solution Description

Introduce a flag to disable the Target Manager if not needed.

_No response_

### Additional Context

End2End testing is quiet complicated due to a default-enabled TM.

puffitos:
## Motivation

Closes #55 

## Changes

- Added if guards for the target manager
- Added simple function to check if a manager is configured
- Set default values for target manager config & sparrowName to `""`, to make the DNS name mandatory and the targetManager optional (this should fix the E2E tests)

## Tests done

- [x] Manual E2E test with a helm install using the parameters used in the E2E workflow succeded.

## TODO


chart/templates/deployment.yaml
```diff
@@ -30,18 +30,19 @@ spec:
...
         - args:
         {{- end }}
           {{- if .Values.startupConfig}}
-          - --config
-          - /startupconfig/.sparrow.yaml
-          - --tmconfig 
-          - /startupconfig/tmconfig.yaml
+            - --config
+            - /startupconfig/.sparrow.yaml
+          {{- else if .Values.targetManagerConfig}}
```
y-eight, line 39: Please use another if here or add the `tmconfig` parameter to the default values in values.yaml `startupConfig` commented out (and no value)

puffitos: I just saw your comment... I had exactly the same though, that was the problem in the first place. I just added 3 ifs, as all cases are currently valid.

The E2E tests pass now.

y-eight (approved):
thx, LGTM

PR #59, opened 2024-01-04
feat: add simple okHandler for api

### LINKED ISSUE #56: Bug: Global Targets need to be handled properly by Checks

### Current Behavior

Currently eg. the health check is using the global targets provided by the target manager directly.
Registered global target `https://sparrow-b.telekom.de` will be used as it is. Every sparrow will provide its health check endpoint at `/checks/health`. The result is that the endpoint seems `unhealthy`.

### Expected Behavior

The sparrow should use the global targets for its checks correctly. Therefore it should append the check's health endpoint (similar to other checks that are using the global targets).
Eg.: `https://sparrow-b.telekom.de/checks/health` 

### Steps To Reproduce

_No response_

_No response_

_No response_

### Anything else?

_No response_

y-eight:
## Motivation

Closes #56 

## Changes

A global handler for "/" that is only returning status ok has been introduced.

## Tests done

- unit test
- local test

### DIFF
--- a/pkg/sparrow/api.go
+++ b/pkg/sparrow/api.go
@@ -59,6 +59,11 @@ func (s *Sparrow) register(ctx context.Context) {
 	// handlers are (de)registered by the checks themselves
 	s.router.HandleFunc("/checks/*", s.handleChecks)
 
+	// Handles requests with simple http ok
+	// Required for global targets in checks
+	s.router.Handle("/", okHandler(ctx))
+
+	// Handles prometheus metrics
 	s.router.Handle("/metrics",
 		promhttp.HandlerFor(
 			s.metrics.GetRegistry(),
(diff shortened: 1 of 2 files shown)


NiklasTreml:
Isn't this already implemented with https://github.com/caas-team/sparrow/pull/57?

y-eight:
#57 is enabling the health check http endpoint as default, rigth!

Targets added by target manager will not have the extension for the check that is using the targets. For health check e.g. `.../checks/health` or latency check `.../v1alpha/latancy`. The easiest way is to expose a global endpoint that can be used as well. Espacially because the checks that are currently available just serve an ok handler as well.
I did not wanted to remove the checks register handler functionylity because we might need it in the future and can be used for manually defined targets.

puffitos (approved):
Funny that the E2E work now 😂

NiklasTreml (approved):

lvlcn-t (approved):

PR #60, opened 2024-01-04
feat: global http client

lvlcn-t:
## Motivation

To enhance the flexibility and maintainability of the HTTP client configuration across various components of the sparrow, I've adopted a context-based approach for managing the HTTP client. This allows the checks, the loader, and the Gitlab manager to customize their client configurations such as timeouts without affecting other parts of the system. This approach also simplifies the Sparrow struct by not including the HTTP client directly, thus adhering to the principles of separation of concerns and reducing coupling.

## Changes

- Implemented a context-based HTTP client management system across all checks, the HTTP loader, and the Gitlab manager.
- Chose not to add the HTTP client directly to the Sparrow struct. This decision was made to keep the Sparrow struct focused on its core responsibilities.
- When a client configuration needs to be changed (e.g., adjusting the timeout), the HTTP client is dereferenced after retrieval from the context. This creates a copy of the client that can be safely modified without causing race conditions or affecting other operations using the client.

For additional information, look at the commits.

## Tests done

- [x] Unit tests succeeded
- [ ] E2E test succeeded

## TODO

### DIFF
--- a/pkg/checks/latency.go
+++ b/pkg/checks/latency.go
@@ -32,6 +32,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/caas-team/sparrow/internal/helper"
+	"github.com/caas-team/sparrow/internal/httpclient"
 	"github.com/caas-team/sparrow/internal/logger"
 	"github.com/caas-team/sparrow/pkg/api"
 )
@@ -44,7 +45,6 @@ func NewLatencyCheck() Check {
 		cfg:     LatencyConfig{},
 		c:       nil,
 		done:    make(chan bool, 1),
-		client:  &http.Client{},
 		metrics: newLatencyMetrics(),
 	}
 }
@@ -54,7 +54,6 @@ type Latency struct {
 	mu      sync.Mutex
 	c       chan<- Result
 	done    chan bool
-	client  *http.Client
 	metrics latencyMetrics
 }
 
@@ -135,13 +134,6 @@ func (l *Latency) SetConfig(_ context.Context, config any) error {
 	return nil
 }
 
-// SetClient sets the http client for the latency check
-func (l *Latency) SetClient(c *http.Client) {
-	l.mu.Lock()
-	defer l.mu.Unlock()
-	l.client = c
-}
-
 func (l *Latency) Schema() (*openapi3.SchemaRef, error) {
 	return OpenapiFromPerfData(make(map[string]LatencyResult))
 }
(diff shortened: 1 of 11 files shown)


lvlcn-t:
I will do the e2e tests tomorrow

pkg/checks/latency.go
```diff
@@ -214,16 +206,26 @@ func (l *Latency) check(ctx context.Context) map[string]LatencyResult {
...
+	// to safely modify its Timeout property for the latency check without affecting other operations.
+	// This is crucial because different checks might require different timeout settings based on
+	// their nature and the targets they're accessing.
+	// By making a copy, we ensure that these modifications are isolated to each specific check,
+	// preventing race conditions and ensuring that each part of the application can specify
+	// its own requirements.
+	// If a check doesn't need to adjust client properties, it could directly retrieve and use
+	// the client from the context just before the request to minimize modifications
+	// and use the default configuration.
+	c := *httpclient.FromContext(ctx)
```
lvlcn-t, line 219: I can remove this comment if you want, my intention with this was to only include it as explanation for the PR

NiklasTreml (approved):

lvlcn-t:
@puffitos I appreciate and share your concern for simplicity and understand your points about the context's intended use.

The motivation behind injecting the HTTP client into the context was to provide a flexible, centralized way to manage HTTP client configurations across all of our components. This approach is particularly relevant where different checks can require varied timeout settings and other configs:

- **Centralization:** A centralized HTTP client reduces the init overhead and potential inconsistency across different checks.
- **Flexibility:** Context configurations allow for per-request adjustments, providing the granularity needed for the checks or even more in the future.

I understand that each check creating and initializing its own HTTP client is a valid approach for simpler scenarios. However, as the sparrow grows with more checks, this could lead to redundancy and complexity:

- Multiple checks mean multiple points of initialization and configuration, leading to code duplication.
- As the number of checks grows, maintaining and updating the HTTP client configurations can become clunky.

---

@NiklasTreml I've considered the utilization of the default client and while it offers simplicity, there are several concerns for our case:

- The default client doesn't easily allow for per-check configurations
- Overwriting a global object can lead to unintended consequences

Nevertheless, I've implemented an alternative approach in the [feat/default-http-client](https://github.com/caas-team/sparrow/tree/feat/default-http-client) branch. This approach utilizes the `http.DefaultClient` to simplify the context and use the default HTTP client.

---

### Alternative Considerations
Instead of overwriting the default client, we can consider other approaches that offer both simplicity and flexibility:

- **Client Pool:** A pool of pre-configured HTTP clients could offer a balance between efficiency and configuration flexibility.
- **Dependency Injection**: We could abstract the HTTP communication behind interfaces and provide concrete implementations at runtime. This offers flexibility and easier testing without compromising simplicity.

Let's find a solution that works well for everyone, catch you next week! 👋

y-eight:
The HTTP client should not be passed in the context in my opinion. Information that is passed along with the context is quite unclear. Active and important components for the functionality should be clearly traceable in the code. Debugging is made more difficult as well.

In terms of centralization & flexibility I agree. A default HTTP client that is used can be good idea (eg. predefined header information, ...)

Lets talk next week. 👌

lvlcn-t:
After our meeting we came to the conclusion that we don't need a global http client anymore.

With this considerations and change of course I've introduced the following pattern:

- If checks require the http client they initialize them for every interval execution
- If other components require a http client with static configurations they initialize them in their struct

I didn't apply the pattern to the health check because I'm doing that in #64.

y-eight (approved):
LGTM

.github/workflows/end2end.yml
```diff
@@ -4,8 +4,11 @@
 name: End2End Testing
 on:
   push:
-    paths:
-    - 'chart/**'
+    branches: 
+      - main
+    tags:
+      - "v[012].[0-9]+.[0-9]+"
+  pull_request:
```
puffitos, line 11: Shouldn't we always run the e2e testing, on each an every push? We need to always make sure that the sparrow can be deployed.

puffitos (changes requested):
I'm not a fan of passing a client as an argument (complicates testing in other languages), as we do in the latency check, but since the `httmock` module has no problems with it, I'm also fine with it :)

.github/workflows/end2end.yml
lvlcn-t: I've added these rules to minimize our CI minutes since we're using the Github Runners.

While we're on the topic, should we use own runners?

puffitos: Total usage of runner in the last 90 days was 33min. We have more than a 1000 each month. Let's keep the E2E tests and I'll open an issue to use a dedicated runner :)

NiklasTreml: I agree with @puffitos. If this becomes relevant we can easily host our own runners. I also talked to to @eumel8, he also doesn't see it as an issue right now

puffitos (approved):

PR #62, opened 2024-01-10
Refactor: Move doc gen from sparrow binary to its own script

NiklasTreml:
## Motivation

The final sparrow binary should not include the gen-docs command, as it is only intended to generate the docs for the repo.
## Changes

Moved the gen-docs into its own package under `scripts/gen-docs`. It can be invoked through `go generate ./...` as before. 

## Tests done

- [ ] Regenerated documentation, `git diff` shows same output


docs/sparrow.md
```diff
@@ -16,7 +16,6 @@ The check results are exposed via an API.
 
 ### SEE ALSO
 
-* [sparrow completion](sparrow_completion.md)	 - Generate the autocompletion script for the specified shell
-* [sparrow gen-docs](sparrow_gen-docs.md)	 - Generate markdown documentation
 * [sparrow run](sparrow_run.md)	 - Run sparrow
 
+###### Auto generated by spf13/cobra on 10-Jan-2024
```
lvlcn-t, line 21: Disable this auto gen tag since the pre-commit hook would always fail once a day.

docs/sparrow_run.md
```diff
@@ -37,3 +37,4 @@ sparrow run [flags]
 
 * [sparrow](sparrow.md)	 - Sparrow, the infrastructure monitoring agent
 
+###### Auto generated by spf13/cobra on 10-Jan-2024
```
lvlcn-t, line 40: Disable this auto gen tag since the pre-commit hook would always fail once a day.

lvlcn-t (changes requested):
One minor thing but after it is fixed: LGTM

puffitos (commented):
Let's remove the date on the auto generation - we can check the date on the modified date of the file :)

scripts/gen-docs/gen-docs.go
```diff
@@ -42,9 +62,10 @@ func NewCmdGenDocs(rootCmd *cobra.Command) *cobra.Command {
...
 
 // runGenDocs generates the markdown files for the flag documentation
-func runGenDocs(rootCmd *cobra.Command, path *string) func(cmd *cobra.Command, args []string) {
-	rootCmd.DisableAutoGenTag = true
+func runGenDocs(path *string) func(cmd *cobra.Command, args []string) {
+	irgendwascmd := sparrowcmd.BuildCmd("")
+	irgendwascmd.DisableAutoGenTag = true
 	return func(cmd *cobra.Command, args []string) {
-		_ = doc.GenMarkdownTree(rootCmd, *path)
+		_ = doc.GenMarkdownTree(irgendwascmd, *path)
```
puffitos, line 69: I would check for the error, we should still check if something goes wrong during doc creation. `irgendwascmd` is funny and I do appreciate a funny name, but I think given the scope of the variable a simple `c` would be enough :)

NiklasTreml: oh yeah forgot to change that :D

lvlcn-t (approved):

y-eight (approved):
LGTM

PR #63, opened 2024-01-10
Refactor: Config and CLI

### LINKED ISSUE #47: Bug: Configuration handling is getting too complicated

### Current Behavior

We should discuss the alternative of how we parse the startup and runtime configuration. Currently, we have to define multiple flags and their string mapping, and we must provide all those long strings in a YAML file, linearly, like this:

```yaml
runParameterOne: 1
mostImporntantRunParameterTwenty: 20
```

This is inflating the codebase and it's questionable whether it's offering us any concrete advantages.

### Expected Behavior

We're able to load the (startup) configuration from one file, which is formatted as follows:

```yaml
checks:
  - check1: {}
  - check2: {}
targetManager: 
  gitlab: {}
# and so on
```

### Steps To Reproduce

_No response_

_No response_

Everyone who has some time to check how the cobra & viper framework works.

### Anything else?

_No response_

NiklasTreml:
## Motivation

The current configuration structure is disorderly, with most options using non-nested, lengthy names. This pull request addresses this issue by reorganizing the configuration layout to utilize nesting, as illustrated below:

Before:
```yaml
fooBarBaz:
fooBarQux:
```
After:
```yaml
foo:
    bar:
        baz:
        qux:
```
This restructuring enhances the clarity and maintainability of the configuration setup.

Adding a parameter to both the configuration and the Command Line Interface (CLI) currently requires changes in three different locations in the codebase. This pull request aims to streamline and simplify this process.

Closes #47

## Changes

Removed flagmapping struct and replaced it with a simple abstraction over Viper and Cobra. The acbstraction provides a unified approach for registering and binding configuration keys to CLI parameters. This abstraction follows a builder pattern, enhancing code structure and maintainability. Also updated helm chart, to reflect the changes to the config.

For additional details, refer to the commit history.

## Tests done
- [x] Updated unit tests to work with the new config


cmd/flag.go
```diff
@@ -0,0 +1,73 @@
...
+type IntFlag struct {
+	f *Flag
+}
+
+type StringPFlag struct {
+	f  *Flag
+	sh string
+}
+
+type BindFN func(cmd *cobra.Command, value, usage string)
```
y-eight, line 26: unused

cmd/flag.go
y-eight, line 1: please add small docu to types and func

cmd/flag.go
```diff
@@ -0,0 +1,73 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+type Flag struct {
+	Config string
+	CLI    string
```
y-eight, line 10: why caps? would prefer `Cli`

chart/values.yaml
```diff
@@ -102,16 +102,22 @@ extraArgs:
...
+  #   http:
+  #     url:
+  #     token:
+  #     timeout:
+  #     retryCount:
+  #     retryDelay:
+  #   file:
+  #     path: /runconfig/checks.yaml
+  # sparrow:
+  #   name: the-sparrow.com
```
y-eight, line 119: it is just `name:` not `sparrow.name:`

pkg/config/config.go
```diff
@@ -104,62 +104,6 @@ func NewConfig() *Config {
...
-}
-
-func (c *Config) SetLoaderFilePath(loaderFilePath string) {
-	c.Loader.file.path = loaderFilePath
-}
-
-// SetLoaderInterval sets the loader interval
-// loaderInterval in seconds
-func (c *Config) SetLoaderInterval(loaderInterval int) {
-	c.Loader.Interval = time.Duration(loaderInterval) * time.Second
```
y-eight, line 128: Please be aware of those conversions!
Currently the time/duration related config parameters are set to seconds in those setter func. 
Same in SetLoaderHttpTimeout & SetLoaderHttpRetryDelay

I would suggest either  (A) switching the flags to `string` type (e.g loader.interval: 3 -> loader.interval: 3s) or (B) convert the duration to seconds in the loader itself.

For A: I am not sure if the viper unmarshaler can unmarshal to time.Duration directly.

y-eight (changes requested):
Please look at the README as well in the end. Docu needs to be changed.
As soon as the auto generation is implemented we might rm the manually written docu.

pkg/config/config.go
y-eight: Sw from setter to the unmarshal is nice! But removing the setter of the config of time & duration will result in a major change in the config as well (old: 3 results in 3 seconds; now: 3 results in 3 nanoseconds).

pkg/config/config.go
```diff
@@ -104,62 +104,6 @@ func NewConfig() *Config {
...
-
-// SetLoaderHttpRetryCount sets the loader http retry count
-func (c *Config) SetLoaderHttpRetryCount(retryCount int) {
-	c.Loader.http.retryCfg.Count = retryCount
-}
-
-// SetLoaderHttpRetryDelay sets the loader http retry delay
-// retryDelay in seconds
-func (c *Config) SetLoaderHttpRetryDelay(retryDelay int) {
-	c.Loader.http.retryCfg.Delay = time.Duration(retryDelay) * time.Second
```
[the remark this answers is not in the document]
NiklasTreml: i have added yaml appropriate yaml and mapstructure tags to the struct where necessary. The yaml parser is already capable to translate strings like `1s` and `1h2m3s` to time.Duration, so we can just use that in the config files. Makes it easier and looks nicer

image

pkg/config/config.go
```diff
@@ -104,62 +104,6 @@ func NewConfig() *Config {
...
-
-// SetLoaderHttpToken sets the loader http token
-func (c *Config) SetLoaderHttpToken(token string) {
-	c.Loader.http.token = token
-}
-
-// SetLoaderHttpTimeout sets the loader http timeout
-// timeout in seconds
-func (c *Config) SetLoaderHttpTimeout(timeout int) {
-	c.Loader.http.timeout = time.Duration(timeout) * time.Second
```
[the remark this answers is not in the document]
NiklasTreml: https://github.com/caas-team/sparrow/pull/63#discussion_r1448949879

pkg/config/config.go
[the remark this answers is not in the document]
NiklasTreml: https://github.com/caas-team/sparrow/pull/63#discussion_r1448949879

NiklasTreml: The parser should fix this on its own: https://github.com/caas-team/sparrow/pull/63#discussion_r1448949879

cmd/flag.go
NiklasTreml: https://github.com/caas-team/sparrow/pull/63/commits/b684ae042a3eab00297836be5c1931de7d79652e adds some docs for the config file

pkg/config/config.go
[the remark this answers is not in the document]
y-eight: perfect 👌

README.md
```diff
@@ -143,6 +143,79 @@ Priority of configuration (high to low):
...
+    # defines which loader to use. Options: "file | http" 
+    type: http  
+    # the interval in which sparrow tries to fetch a new configuration
+    interval: 30s
+    # config specific to the http loader
+    http: 
+        # The url where the config is located
+        url: https://myconfig.example.com/config.yaml
+        # This token is passed in the Authorization header, when refreshing the config
+        token: Bearer xxxxxxx
```
lvlcn-t, line 177: Remove the Bearer since it'd be duplicated if new users keep it and just but their token instead of the placeholder.

The risk is that the authorization header would look like this if they configure it wrong:
```
Authorization: Bearer Bearer xxxx
```

README.md
```diff
@@ -232,11 +305,11 @@ Available configuration options:
...
 - `checks`
   - `latency`
-    - `interval` (integer): Interval in seconds to perform the latency check.
-    - `timeout` (integer): Timeout in seconds for the latency check.
+    - `interval` (duration): Interval to perform the latency check.
+    - `timeout` (duration): Timeout for the latency check.
     - `retry`
       - `count` (integer): Number of retries for the latency check.
-      - `delay` (integer): Delay in seconds between retries for the latency check.
+      - `delay` (furation): Delay between retries for the latency check.
```
lvlcn-t, line 312: typo

pkg/checks/checks.go
```diff
@@ -69,19 +69,19 @@ type Check interface {
...
-	Data any `json:"data"`
+	Data any `json:"data" yaml:"data"`
 	// Timestamp is the UTC time the check was run
-	Timestamp time.Time `json:"timestamp"`
+	Timestamp time.Time `json:"timestamp" yaml:"timestamp"`
 	// Err should be nil if the check ran successfully indicating the check is "healthy"
 	// if the check failed, this should be an error message that will be logged and returned to an API user
-	Err string `json:"error"`
+	Err string `json:"error" yaml:"error"`
 }
```
lvlcn-t, line 78: Why did you apply the yaml tags to the structs even if they have nothing to do with the config?

puffitos:
Everything LGTM @NiklasTreml. I've removed the target manager as well in a separate PR so we can finish this one once and for all. :)

NiklasTreml:
Sounds good 👍 @puffitos

lvlcn-t (approved):
LGTM 👍 but wait with merging until #68 is merged into this branch

cmd/run.go
```diff
@@ -39,88 +39,54 @@ const (
...
+	NewFlag("name", "sparrowName").String().Bind(cmd, "", "The DNS name of the sparrow")
+	NewFlag("loader.type", "loaderType").StringP("l").Bind(cmd, "http", "Defines the loader type that will load the checks configuration during the runtime. The fallback is the fileLoader")
+	NewFlag("loader.interval", "loaderInterval").Int().Bind(cmd, defaultLoaderInterval, "defines the interval the loader reloads the configuration in seconds")
+	NewFlag("loader.http.url", "loaderHttpUrl").String().Bind(cmd, "", "http loader: The url where to get the remote configuration")
+	NewFlag("loader.http.token", "loaderHttpToken").String().Bind(cmd, "", "http loader: Bearer token to authenticate the http endpoint")
+	NewFlag("loader.http.timeout", "loaderHttpTimeout").Int().Bind(cmd, defaultLoaderHttpTimeout, "http loader: The timeout for the http request in seconds")
+	NewFlag("loader.http.retry.count", "loaderHttpRetryCount").Int().Bind(cmd, defaultHttpRetryCount, "http loader: Amount of retries trying to load the configuration")
+	NewFlag("loader.http.retry.delay", "loaderHttpRetryDelay").Int().Bind(cmd, defaultHttpRetryDelay, "http loader: The initial delay between retries in seconds")
```
y-eight, line 57: We still have an issue with the time/delay config here.
The default values do not fit anymore. Eg. defaultLoaderInterval is 300 -> 300ns and should be 300s.
Currently, we are not able to set a default of 300s since this would be an int=300000000000. This is not working for int32.
Any ideas? Should we switch it to type string for time/delay configs?

lvlcn-t: Good catch I think we can just convert the `time.Duration` into int like this:
```go
const (
	defaultLoaderHttpTimeout int = int(30 * time.Second)
	defaultLoaderInterval    int = int(300 * time.Second)
	defaultHttpRetryCount    int = 3
	defaultHttpRetryDelay    int = int(1 * time.Second)
)
```

puffitos: I think we should first try switching to string and seeing if that works, since that's how the durations are parsed in the yaml file. We should refrain from using two conventions for time units, this will complicate our codebase. 

It still bugs me a lot that we're using the `--sparrowName` and the config.yaml value for it is simply `name`.

puffitos: Or would that require extra parsing logic when binding the flag?

y-eight: @lvlcn-t this is not working for int32 i would assume since 300s converted to int will overflow the int32

@puffitos yeah lets sw to `name` should be fine I think

puffitos: There's also a `Duration()` flag, this should solve our problems :) @NiklasTreml  you wanna give a try?

puffitos: @NiklasTreml couldn't resist and started an implementation here https://github.com/caas-team/sparrow/tree/chore/duration-flag, in case you need some inspiration ;)

NiklasTreml: We can just set the flags type to be a `time.Duration` as @puffitos said. 
image
I'll update the flags in the afternoon

y-eight: 😂

puffitos (approved):

y-eight (approved):
Thx! LGTM

lvlcn-t (approved):

Issue #34, opened 2023-12-06
Enhance Health Check Configuration


y-eight:
### Problem Description

The latency check is configurable for several parameters like the interval. The health check should be configurable as well.

### Solution

The following parameters should be configurable:

- interval the check should run
- the http timeout
- the retry count
- the initial retry delay

The implementation can be done similar to the latency check

lvlcn-t:
@y-eight I'd do this issue as part of #33 because I think this should be part of harmonizing the checks.

PR #64, opened 2024-01-10
feat: match up checks & enhance check config

### LINKED ISSUE #33: Match up Health & Latency Checks
# Problem Description

Currently, the two checks (health and latency) have been designed separately from each other. The functionality is similar. The checks architecture should be similar.

Additionally, the latency check should use time references in exposed API keys e.g. if the interval is defined in seconds; use intervalSec: 1.

# Solution

The checks should be matched up so that functionality and code readability is more clear and easier.

lvlcn-t:
## Motivation

To introduce consistency and harmonize existing checks and future implemented checks.

Closes #33 #34 

## Changes

Matched up both existing checks and added the configs mentioned in #34.

## Tests done

- [x] Unit tests succeeded
- [x] E2E tests succeded

## TODO

- [x] Adjust runtime config in helm chart values


lvlcn-t:
Still WIP since I haven't manually E2E tested this yet.

pkg/checks/health.go
```diff
@@ -122,23 +135,30 @@ func (h *Health) Shutdown(_ context.Context) error {
...
+	defer h.mu.Unlock()
+	h.client = c
 }
 
 // Schema provides the schema of the data that will be provided
-// by the heath check
+// by the health check
 func (h *Health) Schema() (*openapi3.SchemaRef, error) {
-	return OpenapiFromPerfData[healthData](healthData{})
+	return OpenapiFromPerfData[[]HealthResult]([]HealthResult{})
```
lvlcn-t, line 161: The current implementation of the health check would change the API schema of the check. 

Should we keep the old schema?

NiklasTreml: I think it's fine if we change this

lvlcn-t: @NiklasTreml Generally what schema would you prefer?
```json
{
  "data": {
    "https://www.google.com/": "healthy",
    "https://www.example.com/": "healthy"
  },
  "timestamp": "2024-01-11T15:19:15.315158452+01:00",
  "error": ""
}
```
or
```json
{
  "data": [
    {
      "target": "https://www.google.com/",
      "status": "healthy"
    },
    {
      "target": "https://www.example.com/",
      "status": "healthy"
    }
  ],
  "timestamp": "2024-01-11T15:19:15.315158452+01:00",
  "error": ""
}
```
The current implementation is like the second schema but I think I'd prefer the first.

NiklasTreml: First one is fine by me

pkg/checks/health.go
```diff
@@ -33,82 +34,88 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-var stateMapping = map[int]string{
-	0: "unhealthy",
-	1: "healthy",
-}
+var (
+	_            Check = (*Health)(nil)
```
y-eight, line 38: Duplicate within health_test.go - witch is the better location I think

y-eight (commented):
pls merge main to branch

pkg/checks/health.go
lvlcn-t: I've moved this from the test file because placing it in the test file could potentially lead to the code compiling successfully even if the `Health` struct doesn't correctly implement the `Check` interface. It also separates the concerns, by keeping production and test code distinct.

chart/templates/configmap.yaml
```diff
@@ -12,6 +12,11 @@ data:
     checks:
       {{- if .Values.runtimeConfig.health}}
       health:
+        interval: {{ .Values.runtimeConfig.health.interval | default 1 }}
+        timeout: {{ .Values.runtimeConfig.health.interval | default 3 }}
+        retry:
+          count: {{ .Values.runtimeConfig.health.retry.count | default 3 }}
+          delay: {{ .Values.runtimeConfig.health.retry.delay | default 1 }}
```
puffitos, line 19: Did you forget the unit here? (1s) There are also 2 other fields with possibly the same problem.

pkg/checks/health.go
```diff
@@ -29,86 +30,90 @@ import (
...
 		}
 	}
 }
 
 // Startup is called once when the health check is registered
-func (h *Health) Startup(_ context.Context, cResult chan<- Result) error {
-	h.c = cResult
+func (h *Health) Startup(ctx context.Context, cResult chan<- Result) error {
+	log := logger.FromContext(ctx).WithGroup("latency")
+	log.Debug("Starting latency check")
```
puffitos, line 114: The health check shouldn't log things about other checks 😜

puffitos (commented):
Everything seems fine, I'll take another look later / tomorrow.

pkg/checks/health.go
puffitos: This should be part of the code being built, as this ensures runtime safety. We should still evaluate if we really need this compile time check, as it may be an indicator that we're adding functions to an interface which are redundant, since we're not using them in our code and we're not offering a library, where not all functions of an interface aren't used by the implementations.

chart/templates/configmap.yaml
lvlcn-t: Ah yes thanks, totally forgot to change that 😄

puffitos (commented):
I've added some minor commits regarding logs & function documentation.

internal/helper/decode.go
```diff
@@ -0,0 +1,81 @@
...
+//	var myConfig MyConfig
+//	myConfig, err := Decode[MyConfig](configMap)
+//	if err != nil {
+//	    // handle error
+//	}
+//
+// This function is particularly useful in settings where the type of the input data might vary,
+// such as when dealing with configuration files in different formats (JSON, YAML, etc.).
+// It provides a convenient and type-safe way to handle such dynamic data.
+func Decode[T any](input any) (T, error) {
```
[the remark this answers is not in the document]
lvlcn-t: I've documented it that detailed because it's a generic function so everyone understands what the it's doing and when it's used. 

But if you say it's only complicating it, then I'll take it out 👍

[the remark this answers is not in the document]
puffitos: It's nice having a clear explanation of what the function does, no questioning that. But one thing is constant in this job; the implementation of things changes over time. Documentation should change then as well. The documentation part doesn't often change when the implementation changes.

This is IMO a generally bad habit of explaining too much. A clear, short and concise explanation of the *behavior* (the most important part of the function) is what we should continually strive for.

puffitos (approved):

pkg/checks/health.go
y-eight: good point

y-eight (approved):

PR #65, opened 2024-01-10
refactor: redistribute checks into smaller packages

lvlcn-t:
## Motivation

The checks package is overloaded and it's getting bigger everytime a new check is introduced.

## Changes

I've redistributed the checks into smaller packages. I needed to create several because of import cycles.

## Tests done

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO


puffitos (commented):
Everything is fine. Some thoughts:

- `oapi` package seems unencessary - it's just a helper function, which could be placed in the top-level `checks` package, as all checks need to create a Schema. It feels wrong having a package about  checks and suddently there's an extra package there regarding open API schema generation?
- The `LatencyConfig`, LatencyResult` and `HealthConfig` can be named to just Config, the package name is already indication enough.
- The NewHealthCheck / NewLatencyCheck can be simplified as well to just `NewCheck`, for the same reason as above (New would be way to abstract, but still fit in the sparrow's context).

lvlcn-t:
@puffitos I've moved the `oapi` helpers into a dedicated package to make the API refactoring easier.
But yes I think we can keep it for now in the `checks` package and move it in the API refactoring PR.

puffitos (approved):

y-eight (approved):

NiklasTreml (approved):

PR #69, opened 2024-01-12
feat: Service Monitor

NiklasTreml:
## Motivation

Since we're exposing metrics for prometheus, we should also provide a ServiceMonitor, so people who are using the prometheus operator can easily monitor their sparrow instance. 

Closes #52 

## Changes
This PR adds a serviceMonitor to the helm chart, which can optionally be enabled.

For additional information look at the commits.

## Tests done
1. Installed sparrow with the default helm chart in a local kind cluster
2. Rolled out kube-prometheus-stack
3. Provisioned a prometheus instance
4. Enabled the serviceMonitor
5. Profit
image
## TODO

- [x] I've assigned this PR to myself
- [x] I've labeled this PR correctly

PR #73, opened 2024-01-16
feat: Prepare helm chart for 0.3.0

NiklasTreml:
## Motivation

Since #69 added servicemonitors to the helm chart, we should also bump the chart version.

## Changes
- bumped chart version to 0.0.4


puffitos:
Can the chart be deployed in its current state?

eumel8:
```
panic: failed to read config file open tmconfig.yaml: no such file or directory
```

so, this is required?

puffitos:
> ```
> panic: failed to read config file open tmconfig.yaml: no such file or directory
> ```
> 
> so, this is required?

We should release 0.0.3 and use that image for the chart. Once #65 is merged, we're good to go.

NiklasTreml:
I've bumped the charts appVersion to v0.3.0 to prepare for release

y-eight (approved):

lvlcn-t:
I've fixed a bug where the yaml couldn't be parsed if the service monitor is enabled but no labels were set.

Now values like this are vaild:
```yaml
# -- Configure a service monitor for prometheus-operator
serviceMonitor:
  # -- Enable the serviceMonitor
  enabled: true
  # -- Sets the scrape interval
  interval: 30s
  # -- Sets the scrape timeout
  scrapeTimeout: 5s
  # -- Additional label added to the service Monitor
  labels: {}
```

lvlcn-t (approved):

NiklasTreml:
please look at #75, as that pr fixes e2e and allows us to properly test 0.3.0 internally

NiklasTreml:
there is still a bug with marshalling env variables correctly, it is semi fixed in viper 1.18.0 but was hidden behind a feature flag in 1.18.2 due to causing a different bug. That bug only appears when using fun stuff like pointer pointers, so we should be fine I think.

https://github.com/spf13/viper/pull/1429
https://github.com/spf13/viper/pull/1715

lvlcn-t (commented):
Shouldn't the chart work out-of-the-box without adjusting any values?

Currently that wouldn't be the case since the name and loader are mandatory:
```bash
$ helm upgrade sparrow chart/ -i --atomic --set image.tag=commit-b1dd2fc

Using config file: /config/.sparrow.yaml
{"time":"2024-01-18T19:08:59.750947469Z","level":"ERROR","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*Config).Validate","file":"/home/runner/work/sparrow/sparrow/pkg/config/validate.go","line":40},"msg":"The name of the sparrow must be DNS compliant"}
{"time":"2024-01-18T19:08:59.751057558Z","level":"ERROR","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*Config).Validate","file":"/home/runner/work/sparrow/sparrow/pkg/config/validate.go","line":47},"msg":"The loader http url is not a valid url"}
Error: error while validating the config: validation of configuration failed

...

error while validating the config: validation of configuration failed
Stream closed EOF for default/sparrow-75d554645d-ms8zq (sparrow)
```
I think we should still include required default configs to improve the user experience with this chart.

lvlcn-t (approved):
LGTM I'll do another PR for docs asap so we can release v0.3.0 today

PR #74, opened 2024-01-17
feat: package helm chart on push

NiklasTreml:
## Motivation

Since we build and publish a docker image for every pr and main push, we should do the same thing for the helm chart.

```bash
❯ helm show chart  oci://ghcr.io/caas-team/charts/sparrow --version 0.2.2-commit-3158f2d
Pulled: ghcr.io/caas-team/charts/sparrow:0.2.2-commit-3158f2d
Digest: sha256:4711bb41a28e0f17ab63faa2890a6c5f4f63822b50413ecaa9be94f23cb86f37

```

```yaml
apiVersion: v2
appVersion: commit-3158f2d
description: A Helm chart to install Sparrow
icon: 'https://github.com/caas-team/sparrow/blob/main/docs/img/sparrow.png'
keywords:
  - monitoring
maintainers:
  - email: f.kloeker@telekom.de
    name: eumel8
    url: 'https://www.telekom.com'
  - email: maximilian.schubert@telekom.de
    name: y-eight
    url: 'https://www.telekom.com'
name: sparrow
sources:
  - 'https://github.com/caas-team/sparrow'
type: application
version: 0.2.2-commit-3158f2d

```

## Changes

Packages the helm chart in the ci workflow on every pr and main commit, and releases it with its version and appVersion set to commit-<GIT_SHORT_SHA>. Ensures that, whenever we build a new docker image, we also have a helm chart available.

## Tests done

- [x] ran pipeline
- [x] `helm upgrade -n sparrow sparrow oci://ghcr.io/caas-team/charts/sparrow --version 0.2.2-commit-3158f2d` resulted in image ghcr.io/caas-team/sparrow:commit-3158f2d being used, as expected

## TODO

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,9 @@ jobs:
         with:
           fetch-tags: true
 
+      - name: List tags
+        run: git tag -l
+
       - name: Get App Version
         id: appVersion
         run:  echo "value=commit-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,9 @@ jobs:
         run: |
           git tag --sort=taggerdate | tail -1 | cut -c 2- 
           echo "value=$(git tag --sort=taggerdate | tail -1 | cut -c 2-)-commit-$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+      - name: DEBUG
+        # produces a tag like: 0.2.2-commit-asdf3
+        run: git tag --sort=taggerdate 
       
       - name: Registry login
         run: helm registry login ghcr.io -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,9 +73,6 @@ jobs:
       - name: Fetch tags explicitly
         run: git fetch --tags
 
-      - name: List remotes
-        run: git remote
-
       - name: List tags
         run: git tag -l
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -61,4 +61,38 @@ jobs:
...
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+
+      - name: Fetch tags explicitly
+        run: git fetch --tags
+
+      - name: List tags
+        run: git tag -l
```
lvlcn-t, line 77: Remove debug step

.github/workflows/ci.yml
```diff
@@ -61,4 +61,38 @@ jobs:
...
+  helm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+
+      - name: Fetch tags explicitly
+        run: git fetch --tags
```
lvlcn-t, line 74: You're probably doing this because of the issue right? 
https://github.com/actions/checkout/issues/1471

NiklasTreml: yep

NiklasTreml:
@y-eight I don't think it's an issue tbh. The helm chart is currently 5.2 KiB small. We could push like a thousand charts for every docker image that we build. Don't think its worth the effort

puffitos:
TLDR; we should only create charts for new, working and tagged releases of sparrow. When we start doing proper E2E testing, we should re-evaluate how we handle charts and container images for usage with E2E testing. We don't really need a new container on each push and we most definetely don't need a new chart.

My two cents:
Even though size isn't an issue now, but management of the github packages leaves much to be desired. I haven't found a quick alternative to being able to clean up the package repository or create automatic cleanup rules. Yeah scripts do exists to do this and Github does have an API, must it's not something we should spend our time doing right now. 

I'm already questioning if it's meaningful to create a container image and pushing it to github for each commit (we only have 500Mi of space for packages and each container is about 12Mi big. We don't even need to push the image to the registry to do the e2e tests (the local image can be pushed into the kubernetes cluster directly).

Do we need this to manually test the helm chart? Since we don't have any test automation, I doubt we're gaining something from having a OCI chart lying around, that we can deploy. I can just build my image locally (it's much quicker) and deploy the current helm chart locally (also quicker).

NiklasTreml:
merging this, but definitely agree with you @puffitos

PR #75, opened 2024-01-18
feat: simplify helm chart

NiklasTreml:
## Motivation

Simplifies the helm chart by removing the separate startup and runtime configs in favour of having a singular sparrow config file.
## Changes

- Merges runtimeConfig and startupConfig into a single object which gets deployed and mounted as a single configmap
- updated github actions to work with new chart

## Tests done
- deployed locally
## TODO

- [ ] I've assigned this PR to myself
- [ ] I've labeled this PR correctly

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -93,13 +93,11 @@ affinity: {}
 
 # -- extra command line start parameters
 # see: https://github.com/caas-team/sparrow/blob/main/docs/sparrow_run.md
-extraArgs:
-  loaderFilePath: /runconfig/checks.yaml
-  loaderType: file
+extraArgs: {}
 
 # -- startup configuration of the Sparrow
 # see: https://github.com/caas-team/sparrow/blob/main/docs/sparrow_run.md
-startupConfig: {}
+config: {}
 # name: the-sparrow.com
 # api:
 #   address:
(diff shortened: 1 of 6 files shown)


y-eight:
My first thought: I dont like that we join the splited config (startup & runtime config), even when using the file loader. This are still different configs. We should prepare the helm chart like this as well.

Second thought: Why not joining, we will just have one config file when e.g. using the file loader

What do you think from user experience point of view? I don't want to confuse joining the config and using different parts of the config by 2 mechanisms... @NiklasTreml @puffitos @lvlcn-t

puffitos:
We should keep the two configurations separate. How the sparrow is setup to run *itself* (what is now called Config or StartupConfig) should have almost nothing to do with how the checks are configured (what is now confusingly called RuntimeConfig and we map to an `any` struct in the code).

Those two configurations should stay separate and be handled in separate ways, starting in the codebase and moving up to the deployment configuration (the helm charts). The sparrow only needs its config to run (the startup config) and if a check configuration file is provided (either via file or via an HTTP endpoint), various checks will also be executed. I'm against providing only one `config` parameter in the helm chart, as this may seem simpler, but we're effectively combining two non-related things. This *will* complicate things down the line.

Let's stick with the 2 configs: `sparrowConfig` and `checksConfig`. Two separate k8s Objects, two separate values and two separate go structs (and interfaces possibly).

NiklasTreml:
The helm chart never had two separate configmaps for the config, but I've update the chart so it now creates one file for each config and loads it into the pod

puffitos:
> The helm chart never had two separate configmaps for the config, but I've update the chart so it now creates one file for each config and loads it into the pod

👍

puffitos (approved):

lvlcn-t (approved):

NiklasTreml:
@puffitos @y-eight extra secrets are finally implemented. took a bit longer than expected because I found a bug with parsing environment variables. Ive fixed the bug in 551f07c03a320265cbb4b1b26995ccc928f02972, but I wish it didn't take me 45 minutes to notice...

chart/values.yaml
```diff
@@ -95,6 +95,15 @@ affinity: {}
 # see: https://github.com/caas-team/sparrow/blob/main/docs/sparrow_run.md
 extraArgs: {}
 
+#
+
+envFromSecrets:
```
lvlcn-t, line 100: add a short comment how to use this/link this https://github.com/caas-team/sparrow/blob/main/README.md#configuration

puffitos:
LGTM (again)

y-eight (commented):
Regarding docu, please update the docu in the main README as well. I would suggest to add a little info how to use a manually created secret in helm context.

I leave it up to you if you want to do that in this PR or open another one.
I can do it as well tomorrow.

Issue #84, opened 2024-01-23
Bug: Nil pointer in http loader if endpoint is not reachable


y-eight:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If the endpoint to load the checks' configuration is not reachable the sparrow will panic with nil pointer exception.

`runtimeCfg` will be nill if the endpoint is not reachable even after retry logic.
https://github.com/caas-team/sparrow/blob/c4a6b545c1b04d871c7e366f1604cc3d36bc1729/pkg/config/http.go#L76

Currently, just a warning log will be printed and the func is not returning/handling the error properly.

### Expected Behavior

Loader should not panic if it is not able to load the remote checks' config.

An else block to cover the none-err case might be a solution.

### Relevant logs and/or screenshots, environment information, etc.

```SH
Using config file: /config/.sparrow.yaml
{"time":"2024-01-23T08:50:09.525650498Z","level":"INFO","source":{"function":"github.com/caas-team/sparrow/cmd.NewCmdRun.run.func1","file":"/home/runner/work/sparrow/sparrow/cmd/run.go","line":82},"msg":"Running sparrow"}
{"time":"2024-01-23T08:50:09.52593781Z","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/sparrow/targets.(*gitlabTargetManager).Reconcile","file":"/home/runner/work/sparrow/sparrow/pkg/sparrow/targets/gitlab.go","line":80},"msg":"Starting global gitlabTargetManager reconciler"}
{"time":"2024-01-23T08:50:09.526036566Z","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/sparrow.(*Sparrow).api.func1","file":"/home/runner/work/sparrow/sparrow/pkg/sparrow/api.go","line":81},"msg":"Serving Api","addr":":8080"}
[...]

Issue #71, opened 2024-01-14
Feature: Shutdown for loaders


puffitos:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

To harmonize the component interfaces of the sparrow and allow for graceful shutdowns, all components should have a shutdown routine integrated.

### Solution Description

Add a `Shutdown(ctx) error` function to the Loader interface and implement it.

### Who can address the issue?

Everyone
cc @lvlcn-t 

### Additional Context

Low level priority task. This is only needed to have a more standardized way of how all sparrow components work. The sparrow should be responsible for shutting down its components, if needed, and the components should always just keep on running as long as they can.

PR #86, opened 2024-01-23
fix/nil-runtime-cfg

puffitos:
## Motivation

- Closes https://github.com/caas-team/sparrow/issues/84
- Closes https://github.com/caas-team/sparrow/issues/71

## Changes

Added the `continue` mentioned in the issue.

To properly test the new changes I had to implement the Shutdown function for the loader, to properly shutdown the loader and not just let the `Run` goroutine end alone.

## Tests done

- New unit tests
- A manual E2E test where I couldn't load the remote RuntimeConfig from Gitlab was also sucessfull -> no panics.

## TODO

### DIFF
--- a/pkg/config/file.go
+++ b/pkg/config/file.go
@@ -33,6 +33,7 @@ var _ Loader = (*FileLoader)(nil)
 type FileLoader struct {
 	path string
 	c    chan<- map[string]any
+	done chan struct{}
 }
 
 func NewFileLoader(cfg *Config, cCfgChecks chan<- map[string]any) *FileLoader {
@@ -61,4 +62,13 @@ func (f *FileLoader) Run(ctx context.Context) error {
 
 	f.c <- cfg.Checks
 	return nil
+}
+
+func (f *FileLoader) Shutdown(ctx context.Context) {
+	log := logger.FromContext(ctx)
+	select {
+	case f.done <- struct{}{}:
+		log.Debug("Sending signal to shut down file loader")
+	default:
+	}
 }
--- a/pkg/config/file.go
+++ b/pkg/config/file.go
@@ -65,6 +65,8 @@ func (f *FileLoader) Run(ctx context.Context) error {
 }
 
 func (f *FileLoader) Shutdown(ctx context.Context) {
+	// proper implementation must still be done
+	// https://github.com/caas-team/sparrow/issues/85
 	log := logger.FromContext(ctx)
 	select {
 	case f.done <- struct{}{}:
--- a/pkg/config/http_test.go
+++ b/pkg/config/http_test.go
@@ -25,8 +25,11 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"sync"
 	"testing"
 	"time"
+
+	"github.com/caas-team/sparrow/internal/helper"
 
 	"github.com/caas-team/sparrow/internal/logger"
 	"github.com/jarcoal/httpmock"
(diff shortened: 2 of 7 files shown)


pkg/config/runtime_config.go
```diff
@@ -19,5 +19,5 @@
 package config
 
 type RuntimeConfig struct {
-	Checks map[string]any `yaml:"checks"`
+	Checks map[string]any `yaml:"checks" json:"checks"`
```
puffitos, line 22: Added the json tag to be able to test using the convenience method httpmock.NewJsonResponder. Else we would have to write a YAML text (which is a hassle to update each time) and use a String Responder each time. I'm open to other suggestions :)

lvlcn-t: It's okay imo, but as a suggestion to split concerns you could write your test input, marshal it to yaml and then give the marshaled yaml to the string responder.
```go
		out, err := yaml.Marshal(tt.response)
		resp := httpmock.NewStringResponder(tt.code, string(out))
```

pkg/config/file.go
```diff
@@ -62,3 +63,14 @@ func (f *FileLoader) Run(ctx context.Context) error {
 	f.c <- cfg.Checks
 	return nil
 }
+
+func (f *FileLoader) Shutdown(ctx context.Context) {
+	// proper implementation must still be done
```
lvlcn-t, line 68: Can you write `TODO:` as comment prefix to alarm IDE extensions about this 😄

lvlcn-t (approved):
LGTM, only minor things that can also be done in the future 👍

pkg/config/http_test.go
```diff
@@ -161,3 +164,237 @@ func TestHttpLoader_GetRuntimeConfig(t *testing.T) {
...
+				defer wg.Done()
+				time.Sleep(time.Millisecond * 600)
+				t.Log("Shutting down the Run method")
+				hl.Shutdown(ctx)
+			}()
+
+			err := hl.Run(ctx)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("HttpLoader.Run() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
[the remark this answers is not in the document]
puffitos: I'll try. I battled a lot with the tests till I got them running without any race conditions. 1 min for the change, 1h 30m for the ~~tears~~ tests.

pkg/config/http_test.go
```diff
@@ -161,3 +164,237 @@ func TestHttpLoader_GetRuntimeConfig(t *testing.T) {
...
+	case <-time.After(time.Second):
+		t.Error("Config not sent to channel")
+	case c := <-cCfgChecks:
+		if !reflect.DeepEqual(c, expected) {
+			t.Errorf("Config sent to channel is not equal to expected config: got %v, want %v", c, expected)
+		}
+	}
+
+	hl.Shutdown(ctx)
+	httpmock.DeactivateAndReset()
```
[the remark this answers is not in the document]
puffitos: I'll do this tomorrow, this is a leftover artifact before I understood how to make the tests run without triggering the race conditions

pkg/config/runtime_config.go
puffitos: Sure, that sounds more than reasonable 😅

puffitos: I used the ByteResponder, to avoid casting to string.

pkg/config/http_test.go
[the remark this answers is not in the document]
puffitos: I've added the httpmock.Reset(); I don't think it's necessary. The whole race-condition-fiasco happens because the blocking routines (like the checks checking a remote endpoint or the loader trying to grab the config from a remote endpoint) are still getting executed, when the tests end and their `Run`  routine is still running (it was never properly terminated). httpmock tries to deactivate the mocks by accessing the transport, but it's still under use from the `Retry` function -> race condition.

y-eight (approved):
LGTM, thx!

Issue #80, opened 2024-01-19
Prune old images and charts


lvlcn-t:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

The github container registry is getting/will get really big because we're pushing all commit images and charts into it.

In order not to lose clarity we should prune old images and charts.

![image](https://github.com/caas-team/sparrow/assets/75443136/c4325d4a-40d9-4327-9eb7-3b9087ba140d)![image](https://github.com/caas-team/sparrow/assets/75443136/8cf5f4dc-8384-4c8f-80fe-048d49362935)

### Solution Description

Add a new github action that runs periodically that prunes all images older than 7 days or so.

I've found a github action that can do this, but it's unmaintained. I've linked it below.

Additionally we should think about only packaging and pushing the charts if they've changed.

### Who can address the issue?

Everyone who wants to explore github actions.

### Additional Context

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
https://github.com/vlaurin/action-ghcr-prune

PR #88, opened 2024-01-25
feat: added new workflow to prune old images

JTaeuber:
## Motivation

Relates to https://github.com/caas-team/sparrow/issues/80
## Changes
Added a new workflow to prune old images of packages sparrow and charts/sparrow.

The workflow is configured to prune all images older than 7 days that start with 'commit-' or end in 'SNAPSHOT-...' for the sparrow package. 

For the charts/sparrow package it will prune all images older than 7 days that end in 'commit-...'.

It runs daily at midnight (could be changed if wanted).

It is still configured to dry-run just to be sure. This should be changed once we're sure nothings going to break.

## Tests done
I've tested the workflow in dry-run in a fork I made and it worked accordingly.

No stable version or version under 7 days was marked for pruning.

https://github.com/JTaeuber/sparrow/actions/runs/7642146860
This is the Test of the Workflow without keeping the images younger than 7 days for the charts/sparrow package, so you can see what it would prune if they were older than 7 days.

## TODO

### DIFF
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -48,8 +48,6 @@ jobs:
           keep-younger-than: 7 # days
           keep-last: 2
           prune-untagged: true
-          keep-tags: |
-            0.0.1
-            0.0.2
-            0.0.3
-            0.0.4
+          prune-tags-regexes: |
+            ^0.2.2-commit
+            ^0.3.0-commit
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -29,7 +29,6 @@ jobs:
           container: sparrow # Package name
           dry-run: true # Dry-run first, then change to `false` or remove
           keep-younger-than: 7 # days
-          keep-last: 2
           prune-untagged: true
           prune-tags-regexes: |
             ^commit-
@@ -43,7 +42,6 @@ jobs:
           container: charts/sparrow # Package name
           dry-run: true # Dry-run first, then change to `false` or remove
           keep-younger-than: 7 # days
-          keep-last: 2
           prune-untagged: true
           prune-tags-regexes: |
             commit-.*$
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -49,4 +49,4 @@ jobs:
           keep-last: 2
           prune-untagged: true
           prune-tags-regexes: |
-            'commit-*$'
+            '(?<=:\s*).*\bcommit\b.*'
--- a/.github/workflows/prune.yml
+++ b/.github/workflows/prune.yml
@@ -11,7 +11,14 @@ permissions:
 
 jobs:
   prune_images:
-  
+    name: Prune old sparrow images
+    runs-on: ubuntu-latest
+
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/prune.yml
```diff
@@ -0,0 +1,47 @@
+name: Prune GHCR
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Runs workflow at '(minute) (hour) (day of the month) (month) (day of the week)'
```
y-eight, line 5: Pls either rm the comment or state the actual behavior.

.github/workflows/prune.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    steps:
+    
+      - name: Prune Images
+        uses: vlaurin/action-ghcr-prune@v0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          organization: caas-team
+          container: sparrow # Package name
+          dry-run: true # Dry-run first, then change to `false` or remove
```
y-eight, line 30: rm comment

.github/workflows/prune.yml
```diff
@@ -0,0 +1,47 @@
...
+      security-events: write
+
+    steps:
+    
+      - name: Prune Images
+        uses: vlaurin/action-ghcr-prune@v0.5.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          organization: caas-team
+          container: sparrow # Package name
```
y-eight, line 29: rm comment

y-eight (approved):
LGTM

.github/workflows/prune.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  prune_images:
+    name: Prune old sparrow images
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      packages: write
+      security-events: write
```
lvlcn-t, line 20: These permissions aren't necessary since you've declared them as default permissions above.

Ref: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

.github/workflows/prune.yml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          organization: caas-team
+          container: sparrow
+          dry-run: true
+          keep-younger-than: 7 # days
+          prune-untagged: true
+          prune-tags-regexes: |
+            ^commit-
+            SNAPSHOT-.*$
```
lvlcn-t, line 35: After the first prune we can remove this regex because our new images don't follow this naming pattern anymore.

puffitos: The regex is unnecessary. We just want to remove old images, no matter what naming convention they used. Let's just remove it altogether.

lvlcn-t: Then we'd also remove our stable images because git tags != image tags

puffitos: Totally right, my bad 😅

puffitos (approved):

Issue #72, opened 2024-01-16
Refactor: remove any typed Configuration


puffitos:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

We're using `any` as a type in multiple places to parse the Runtime Configuration of the checks. Since we're not actually using such a feature, we should move away from using the any interface. It complicates our code and makes parsing configuration for Checks prone to errors.

### Solution Description

- Use a RuntimeConfiguration struct to handle the configuration for all checks (as done now) and also use this in all places where the RuntimeConfiguration is touched upon (like the reconciling of the config itself), so we use something that's as close to universal as possible.
- This object will be passed around and the per-check Config can be used in the needed places, with functions like GetHealthConfig().

PR #93, opened 2024-01-27
refactor/typed-runtime-cfg

puffitos:
## Motivation

Closes <https://github.com/caas-team/sparrow/issues/72>

Since we currently have no binding reasons of using `any` typed configurations if a strongly typed language, we should avoid doing that :) Please read through the text of the PR and come back to it. I've tried to explain everything done. 

It's _only_ like 1000 lines of code changed :+1: Although I encourage organizing a call to go over the PR, please go through the changes yourselves; most of the changes are just renamings of imports... The important parts are the interface changes and mostly the ReconcileChecks function of the Sparrow. The rest was just doing due diligence and updating everything so that the code may compile.

Needless to say, to be merged after #91. Also includes most changes from that PR. 

### Any type removal

The `any` Configuration was removed in the following fashion:

- A new struct named `Config`, that represents the YAML file with the check’s runtime configuration was introduced.
- The `Config` struct was used everywhere, where the `map[any]any` appeared.
- The general and abstract character of the `any` map can’t be expressed directly by a typed `Config` - this created the need for an interface for those CheckConfigs coming in each time to be able to differentiate between them, without doing type checking each time.
- Additionally, the Checks themselves needed a way to manage their configurations during the ReconcileChecks function of the sparrow. This raised the need for two new interface functions for the Checks themselves, to be able to identify them and manipulate their configs (`Name` and `GetConfig`).
- The check’s `Config` structs must thus implement a new interface, currently named `Runtime` (used as `checks.Runtime` - I’m open to better names. The name occurred “organically” from `RuntimeConfig`, which is the current type being used in the `config` package). This handles the checks and their configuration during the sparrow’s runtime and reconciliation methods.

The rest was:

- change each Check’s `SetConfig` and implement the `Name` and `GetConfig`
- refactor the sparrow’s methods in `run.go` - this was the most difficult part, as the ReconcileChecks function is heavy. We should see how we can simplify the logic, if that’s possible.
- fix tests that used the any maps - that was just painful.

### Structural

Some moving around of packages had to be done, because of cyclic imports. We should generally avoid creating dedicated packages up until the point we realize, that a package is too big and handles too many responsibilities.

Changes:

- `types` package is gone, contents were moved to `checks` into a `base.go` file, which contains the basic types for the checks and their results.
- `errors` package was unneeded, can be part of the `checks` package. Also, naming packages after the stdlib is something we should avoid.
- The `register.go` file was replaced by an sister-to-`checks` package named `factory`, to give more meaning to the existence of this “registry” for checks. A factory to create the checks, given a runtime configuration. The management of which checks are available and their creation shouldn’t be part of the `checks` package, but a package which *uses* the `checks` package.
- A `runtime` package in `checks` was introduced, to handle the Runtime Configuration of the checks. It currently cannot be part of the `checks` package, as the checks themselves need this package to run. Maybe it’s place is in a different package altogether, or maybe we shouldn’t package the checks 🤷

---

**You may stop reading here and just start the review. If you need more information though, keep reading on the specific changes.**

---

### Checks runtime configuration

This struct handles the runtime configuration file for all checks. This struct replaces the `any` maps with the checks configuration in the whole codebase.

Notable features:

- Every time a check is added, we *must* update the `Config` struct as well to handle the extra configuration for the check, or else the check won’t be configurable.
- The struct has some convenience functions which allow getting the whether the config is empty, whether checks exist and iterating over the checks’ configurations. Those must also be updated for each new check and are mostly used in the already present functions.

The runtime configuration of the checks was removed from the startup configuration check struct (in the `config` package), as this was the wrong place for it. Finally, the unused and currently unneeded `apiVersion` and `kind` fields of the configuration were removed from the configuration file. RIP (not).

Still open (for the future):

- The Interface should also have a `Validate` method to validate the configuration per check. Left it out, as this would only enlarge this already large enough PR.
- The Validate can then trickle up to the `runtime.Config` struct, which should have a validate Method to validate all configurations.

### Check Interface & Check handling

The Check interface was modified, as it was the most sensible thing that came in mind to be able to handle the update of the checks themselves during the `ReconcileChecks` routine. The interface was thus expanded:

- to return the `Name` of the Check, which allows the identification of the check
- to be able to get the current Config of the check, to handle the update of the configuration of the current checks. This seems a bit weird, maybe we should fix this.

To create new checks, a proper Factory was created to just call the constructors and also create multiple checks from the incoming `runtime.Config` from the configuration channel of the Sparrow.

## Tests done

- Unit tests run
- Simple E2E tests run (we must run E2E tests ASAP)

## Final remarks

Most fragile changes of this refactoring were made because of how the `ReconcileChecks` works, like expanding the Checks interface with a `GetConfig`, which seems unnecessary to just update the config of a check. We should focus some time on finding a simpler logic on how to CRUD the checks managed by the sparrow, instead of managing them in a `map[string]checks.Check` (a map of names to *interfaces*).


pkg/config/http.go
```diff
@@ -25,22 +25,24 @@ import (
...
 	"gopkg.in/yaml.v3"
 )
 
 type HttpLoader struct {
-	cfg        *Config
-	cCfgChecks chan<- map[string]any
-	done       chan struct{}
-	client     *http.Client
+	cfg         *Config
+	cConfigCfgs chan<- runtime.Config
```
lvlcn-t, line 37: Rename to `cRuntime` or something more meaningful

pkg/config/http_test.go
```diff
@@ -294,19 +271,20 @@ func TestHttpLoader_Run_config_sent_to_channel(t *testing.T) {
...
+			Interval: 1 * time.Second,
 		},
 	}
-	body, err := yaml.Marshal(RuntimeConfig{Checks: expected})
+	body, err := yaml.Marshal(expected)
 	if err != nil {
 		t.Fatalf("Failed marshaling yaml: %v", err)
 	}
 	resp := httpmock.NewBytesResponder(200, body)
 	httpmock.RegisterResponder("GET", "https://api.test.com/test", resp)
```
lvlcn-t, line 285: `http.MethodGet`

pkg/sparrow/run.go
```diff
@@ -137,120 +140,93 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
-	if s.tarMan == nil {
-		return checkCfg
-	}
 	gt := s.tarMan.GetTargets()
 
-	// filter out globalTargets that are already in the config and self
-	for _, t := range gt {
-		if slices.Contains(actual, t.Url) {
+	// merge global targets with health check targets
+	for _, gt := range gt {
```
lvlcn-t, line 198: rename the slice to avoid variable shadowing:
```go
for _, gt := range gts {
```

pkg/sparrow/run_test.go
```diff
@@ -267,108 +223,118 @@ func TestSparrow_Run_ContextCancel(t *testing.T) {
 	time.Sleep(time.Millisecond * 30)
 }
 
-// TestSparrow_updateCheckTargets tests that the updateCheckTargets method
-// updates the check targets, if they exist in the config of the checks.
-func TestSparrow_updateCheckTargets(t *testing.T) {
+// TestSparrow_enrichHealthTargets tests that the enrichTargets method
+// updates the HealthCheck targets, if they exist in the config of the checks.
+func TestSparrow_enrichTargets(t *testing.T) {
```
lvlcn-t, line 228: godoc name != test function name

pkg/sparrow/run.go
```diff
@@ -137,120 +140,93 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
 
-	if s.tarMan == nil {
-		return checkCfg
-	}
 	gt := s.tarMan.GetTargets()
 
-	// filter out globalTargets that are already in the config and self
-	for _, t := range gt {
-		if slices.Contains(actual, t.Url) {
+	// merge global targets with health check targets
```
lvlcn-t, line 197: _merge global targets with check targets_

lvlcn-t (commented):
First review, I'll look at it again probably on Monday and we should definitely setup a call for this PR next week, it's quite big.

pkg/checks/latency/latency_test.go
```diff
@@ -159,12 +158,12 @@ func TestLatency_Run(t *testing.T) { //nolint:gocyclo
...
-			result := <-results
+			res := <-results
 
-			assert.IsType(t, tt.want.Data, result.Data)
+			assert.IsType(t, tt.want.Data, res.Data)
 
-			got := result.Data.(map[string]Result)
-			expected := result.Data.(map[string]Result)
+			got := res.Data.(map[string]result)
+			expected := res.Data.(map[string]result)
```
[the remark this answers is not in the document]
puffitos: Done. Removed the frail error string checking test.

go.mod
```diff
@@ -6,13 +6,14 @@ require (
 	github.com/getkin/kin-openapi v0.120.0
 	github.com/go-chi/chi/v5 v5.0.10
 	github.com/jarcoal/httpmock v1.3.1
-	github.com/mitchellh/mapstructure v1.5.0
 	github.com/spf13/cobra v1.8.0
-	github.com/spf13/viper v1.18.0
+	github.com/spf13/viper v1.18.2
```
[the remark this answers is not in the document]
puffitos: I'll take a look at that the next days, but running with the normal yaml config works. Viper is doing all sorts of weird things with the multiple configs and parsing options we have... I'm wondering if we really need cobra and viper to just read a yaml file and merge it with environment variables.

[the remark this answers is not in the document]
puffitos: Environment variables are being parsed correctly and overwriting values found in the read in YAML config fine, even with the latest viper version. Let's talk about viper and cobra as well next week and reevaluate them.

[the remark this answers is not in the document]
NiklasTreml: The feature is hidden behind a feature flag. We can use this, but we need to update the builds and tests, to include the build tags

https://github.com/spf13/viper/commit/ab3a50c0ce2020028f8cb515c62fc4d2d97a8996

pkg/sparrow/run.go
```diff
@@ -137,118 +140,89 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
 			continue
 		}
-		if t.Url == fmt.Sprintf("https://%s", s.cfg.SparrowName) {
-			continue
+		if cfg.HasHealthCheck() && !slices.Contains(cfg.Health.Targets, gt.Url) {
+			cfg.Health.Targets = append(cfg.Health.Targets, gt.Url)
+		}
+		if cfg.HasLatencyCheck() && !slices.Contains(cfg.Latency.Targets, gt.Url) {
+			cfg.Latency.Targets = append(cfg.Latency.Targets, gt.Url)
 		}
```
lvlcn-t, line 204: You've consciously didn't enrich the dns ckecks' targets?

pkg/checks/dns/dns.go
```diff
@@ -119,31 +136,24 @@ func (d *DNS) Shutdown(_ context.Context) error {
...
-	}
-
-	// To be able to lookup global targets injected to config targets
-	for k, v := range c.Targets {
-		c.Targets[k], _ = strings.CutPrefix(v, "https://")
+func (d *DNS) SetConfig(cfg checks.Runtime) error {
+	if c, ok := cfg.(*Config); ok {
+		for k, v := range c.Targets {
+			c.Targets[k], _ = strings.CutPrefix(v, "https://")
+		}
```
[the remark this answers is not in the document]
puffitos: Not anymore. I think the checks should just be able to go on about their lives and just execute, without having to care too much about configuration. The targetManager should do the "heavy lifting" of adapting itself to the checks. The other way around would build a dependency of the checks to the targetManager component, and this is a can of worms we don't want to open.

pkg/sparrow/run.go
puffitos: I didn't get around to do it up until now. Thanks for pointing this out! It'n now implemented and tested.

pkg/sparrow/run.go
```diff
@@ -137,118 +141,93 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
-		log := logger.FromContext(ctx).With("name", name)
-
-		c := s.updateCheckTargets(checkCfg)
-		if existingCheck, ok := s.checks[name]; ok {
-			// Check already registered, reset config
-			err := existingCheck.SetConfig(ctx, c)
+func (s *Sparrow) ReconcileChecks(ctx context.Context, cfg runtime.Config) {
+	// generate checks from configuration
+	s.enrichTargets(cfg)
+	fChecks, err := factory.NewChecksFromConfig(cfg)
```
lvlcn-t, line 147: The check's targets are never enriched since you've forgot use the return value of `enrichTargets(cfg)`

lvlcn-t:
@puffitos Can you create an issue to the validation part you've mentioned in your PR?

lvlcn-t (changes requested):

pkg/sparrow/run.go
lvlcn-t: Additionally write a unit test for this in the reconcile testing

y-eight (commented):
Regarding the `Runtime` interface name, no better idea for now. Seems ok for me for now!

pkg/checks/runtime/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+
+// HasLatencyCheck returns true if the check has a latency check configured
+func (c Config) HasLatencyCheck() bool {
+	return c.Latency != nil
+}
+
+// HasDNSCheck returns true if the check has a dns check configured
+func (c Config) HasDNSCheck() bool {
+	return c.Dns != nil
+}
```
NiklasTreml, line 85: Refactor this to call HasCheck() internally

go.mod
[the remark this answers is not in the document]
puffitos: Rolled back to 1.18.0 for the time being.

pkg/checks/errors.go
```diff
@@ -16,8 +16,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package errors
+package checks
 
 import "errors"
 
-var ErrInvalidConfig = errors.New("invalid config")
+var ConfigMismatch = errors.New("the given config does not match the check's type")
```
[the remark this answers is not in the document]
puffitos: Added a dedicated error type to end ambiguity concerns.

pkg/factory/factory.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/caas-team/sparrow/pkg/checks/dns"
+
+	"github.com/caas-team/sparrow/pkg/checks"
+	"github.com/caas-team/sparrow/pkg/checks/health"
+	"github.com/caas-team/sparrow/pkg/checks/latency"
+	"github.com/caas-team/sparrow/pkg/checks/runtime"
+)
+
+// New creates a new check instance from the given name
+func New(cfg checks.Runtime) (checks.Check, error) {
```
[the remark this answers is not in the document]
puffitos: Well said, this remained exposed as it was first called multiple times in the ReconcileChecks function and was there before the convenience, multi creator function.

pkg/sparrow/run.go
puffitos: Great catch, I've added the a mock gitlab target manager to the run_test + a DNS test

pkg/checks/runtime/runtime.go
puffitos: Don't you think that this causes an unneeded indirection? I'm happy that we have dedicated functions that answer the question "is the check configured?". The HasCheck is a helper function only needed for the ReconcileChecks... 

I'm also not happy with it's existence, but I wouldn't move the logic for each check (currently the same for each check) in that function, as it isn't its place IMO. Instead I would revisit the ReconcileChecks function in the future and see whether we really need the `HasCheck` function.

pkg/sparrow/run.go
```diff
@@ -137,118 +141,93 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
 	}
-	var urls []string
+}
 
-	if s.tarMan == nil {
-		return checkCfg
+// enrichTargets updates the targets of the sparrow's checks with the
+// global targets. Per default, the two target lists are merged.
+func (s *Sparrow) enrichTargets(cfg runtime.Config) runtime.Config {
+	if cfg.Empty() || s.tarMan == nil {
```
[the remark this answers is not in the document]
puffitos: Since there currently is no validation (outside the YAML syntax one) when grabbing a new `runtime.Config`, the following config may also be pushed to the channel:

```
incoming := runtime.Config{}

// Parsed from an empty YAML file, which is a legit option
```

We should either stop something from this ever happening at loader-time, or just write a simple guard here, which makes more sense to me.

The point about reworking the function is valid.

lvlcn-t (approved):
LGTM but can you open the validation issue I mentioned before and maybe an issue to your final remark regarding the mapping of check names to interfaces.

pkg/checks/runtime/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+		size++
+	}
+	if c.Latency != nil {
+		size++
+	}
+	if c.Dns != nil {
+		size++
+	}
+	return size
+}
```
NiklasTreml, line 70: Use the corresponding HasXxx function instead:
```go
if c.HasHealthCheck() {
  size++
}
```

puffitos: Good point, will do

NiklasTreml (approved):

PR #61, opened 2024-01-05
Error handling; don't panic.

puffitos:
## Motivation

Addresses #53

Still a WIP. I just wanted to put everything together as documentation so we can discuss this together next week.

## Changes

A friendly guide for the changes:

- No panic, only errors
- Two new channels for sparrow, to handle errors and to handle end of life
- All sparrow components start as goroutines in `Run` and write their errors into the error channel (when something irrecoverable happens)
- The errors are handled by a separate goroutine (`handleErrors`), which may be extended to handle errors in other ways (currently only fatal errors are expected in the error channel)
- The `handleErrors` will gracefully shutdown all sparrow components (if possible) and return the error(s) which led to the shutdown, along with other errors that happened during shutdown.
- Each component should/ must have a shutdown function, to be terminated gracefully

Additionally, I couldn't resist the urge and did the following (sorry!):

- Added the source to the logger, so we know where the log is produced
- Removed the bulky and unhelpful `withGroup` from all loggers

## Tests done

- [x] Normal run without interruptions with local config
- [x] Normal run without interruptions with remote config
- [x] Running with a context that will only last 30 seconds -> routines shutdown and sparrow exits with 1
- [x] Running with misconfigured sparrow config (wrong filepath) -> api starts and shuts down immediately → sparrow exits with 1

### Normal run, metrics from remote config

```
# HELP sparrow_health_up Health of targets
# TYPE sparrow_health_up gauge
sparrow_health_up{target="https://caas-max-sparrow.caas-t02.telekom.de"} 1
sparrow_health_up{target="https://caas-max-sparrow.caas-t02.telekom.de/checks/health"} 1
sparrow_health_up{target="https://caas-max-sparrow.caas-t21.telekom.de"} 1
sparrow_health_up{target="https://gitlab.devops.telekom.de"} 1
sparrow_health_up{target="https://www.google.com/"} 1
# HELP sparrow_latency_count Count of latency checks done
# TYPE sparrow_latency_count counter
sparrow_latency_count{target="https://caas-max-sparrow.caas-t02.telekom.de"} 15
sparrow_latency_count{target="https://caas-max-sparrow.caas-t21.telekom.de"} 15
sparrow_latency_count{target="https://example.com/"} 14
sparrow_latency_count{target="https://gitlab.devops.telekom.de"} 14
sparrow_latency_count{target="https://google.com/"} 14
sparrow_latency_count{target="https://yam.telekom.de"} 14
# HELP sparrow_latency_duration Latency of targets in seconds
[...]

PR #95, opened 2024-01-30
refactor: cleanup and move API 

### LINKED ISSUE #92: Cleanup API with dynamic Handler
With #91

lvlcn-t:
## Motivation

To finish the refactoring started in #91, where the checks' handler methods were removed from the interface.

Closes #92 

## Changes

- I've refactored the API to a dedicated component in another package behind an interface. We partly discussed this [here](https://github.com/caas-team/sparrow/pull/61#discussion_r1451672721).
- I've refactored the `Sparrow` struct into a composition structure in order to improve code organization and clarity. This includes setting apart different responsibilities into `checkCoordinator` and `errorHandler` types.
- I've removed functionalities and sections of code that were no longer needed.

(I also adjusted the docs of one dns metric that has changed)

## Tests done

- I've moved/refactored the unit tests for the API
- I've added some more unit tests
- [x] Unit tests suceeded.
- [x] E2E tests succeeded.

## TODO


pkg/api/api.go
```diff
@@ -0,0 +1,213 @@
...
+	Extensions: make(map[string]any),
+	Components: &openapi3.Components{
+		Schemas: make(openapi3.Schemas),
+	},
+	Servers: openapi3.Servers{},
+}
+
+// GenerateCheckSpecs generates the OpenAPI specifications for the given checks
+// Returns the complete OpenAPI specification for all checks
+func GenerateCheckSpecs(ctx context.Context, cks map[string]checks.Check) (openapi3.T, error) {
```
lvlcn-t, line 183: I don't know about having this here (old function was named [Openapi](https://github.com/caas-team/sparrow/blob/main/pkg/sparrow/api.go#L150)).

Maybe it'd be better to move it here https://github.com/caas-team/sparrow/blob/2ff55bcd9d345829ff555731cd03ed66bdb57925/pkg/sparrow/handlers.go#L39 because it's only used there.

pkg/api/api.go
```diff
@@ -0,0 +1,213 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/caas-team/sparrow/internal/logger"
+	"github.com/caas-team/sparrow/pkg/checks"
+	"github.com/caas-team/sparrow/pkg/config"
```
puffitos, line 30: I find this import weird, right? Why does the API have to import its own config from someplace else. I guess if we move the config in the api package we're gonna have cyclic imports. Pff, that just seems wrong :D

Just something to think about.

puffitos (changes requested):
All in all, everything seems fine. The `Run` method of the sparrow has to change, to keep the focus on its responsibility.

pkg/sparrow/run.go
```diff
@@ -116,8 +113,31 @@ func (s *Sparrow) Run(ctx context.Context) error {
...
+					promhttp.HandlerOpts{Registry: s.metrics.GetRegistry()},
+				).ServeHTTP,
+			},
+		}
+		err := s.api.RegisterRoutes(ctx, routes...)
+		if err != nil {
+			log.Error("Error while registering routes", "error", err)
+			s.cErr <- err
+		}
+		s.cErr <- s.api.Run(ctx)
```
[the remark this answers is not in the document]
lvlcn-t: For now I've just moved the API startup from the `Run` method to another method of the sparrow because the `/openapi` handler currently needs the sparrows check map to properly generate the check oapi specs during runtime and it'd feel awkward to have multiple places for the handlers.

This will also make the refactoring of the sparrow's database responsibility a bit tricky so therefore it should be done with/after #98

pkg/sparrow/run.go
```diff
@@ -27,78 +27,75 @@ import (
...
-
-	cfg *config.Config
-
 	// cRuntime is the channel where the loader sends the checkCfg configuration of the checks
 	cRuntime chan runtime.Config
 	// cResult is the channel where the checks send their results to
 	cResult chan checks.ResultDTO
+}
+
+type errorHandler struct {
```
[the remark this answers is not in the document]
lvlcn-t: It really comforts me to have this (the sparrow struct was an eyesore for me since #5) and imo it makes the struct more readable but if you feel like this is unnecessary/complicating things I'll revert it.

pkg/api/api.go
lvlcn-t: Nope we don't, I've moved the config to the package.

pkg/sparrow/run.go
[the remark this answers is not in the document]
puffitos: No need to revert, let's just roll with it. I do share your concern and thankfully the change is easy to revert or refactor.

puffitos (approved):

pkg/api/api.go
lvlcn-t: I'll move it to the check controller with #98

Issue #36, opened 2023-12-06
Bug: A check that fails to run when initialized won't run at all


puffitos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, during the reconciliation of checks, if a new config enters the `cCfgChecks` channel:

```go
// after checking if the check is already registered (if yes, it is skipped), the check is carried out

		go func() {
			err := check.Run(ctx)
			if err != nil {
				log.ErrorContext(ctx, "Failed to run check", "name", name, "error", err)
			}
		}()
```

if an error is thrown, then the check won't run at all, as it is already registered.

### Expected Behavior

A check that fails to run after being initialized, should still try to run again.

### Steps To Reproduce

It's complicated to reproduce this. A simple unit test could be written, with a check that has a Run method that simply errors out. The test should wait for the checks retry/ delay window, and see if any results with errors are coming in.

Issue #43, opened 2023-12-14
Bug: Panic during Check Shutdown


y-eight:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, the Sparrow is shutting don the fanIn data channels for the checks itself.
When shutting down a check that is currently running, the result data channel might be already closed. This is resulting in a panic while the check tries to write in the result channel.down

### Expected Behavior

A soft shutdown should be done. The check might close the resultData fanIn channel itself. 
The check should not write in a closed channel to be sure not to panic.

PR #102, opened 2024-02-08
refactor: check reconcilation

### LINKED ISSUE #98: Refactoring: Check reconciliation simplification

### Problem Description

The check (config) reconciliation logic should be simplified. We've currently employed a map of check names to checks, which we are exclusively using to register, delete or update the checks. We're using this map structure, to handle more complicated logic like:

- if the map is empty, register all checks
- iterate over the map and start unregistering checks, in case one of the names is not in the incoming config
- iterate over the created checks (from the incoming config) and update/ register the checks, if they exist in the map.

Even though we are not misusing the data structure, it doesn't make for the most simple experience to have to deal with a map. This makes the whole reconciliation logic (the triplet of functions - `ReconcileCheck`, `registerCheck` and `unregisterCheck`) a tad complicated.

### Solution Description

The sparrow is currently in charge of actively reconciling the configurations of the various checks, when a new runtime.Config appears in its runtime channel.

There are multiple ways we could handle this, but the creation of a dedicated struct which will handle the Checks seems like the way to go. A `ChecksController` if you may, which will handle the reconciliation logic of the Checks (registration, update and removal), depending on the runtime configuration channel. Given the modular approach of the sparrow, this feels like the proper way to go; another component of the struct, which can start, act and shutdown autonomously (like the API, the `TargetManager`, the `Loader` and even the `Check` instances themselves).

A more simple solution would be to just create a simple struct to replace the map[string]Check and add some business logic to it (i.e with `updateConfigForCheck(name string, cfg checks.Config).

The end result should make the registration, update and removal of checks a simple affair, which is easy to read and interpret and, most importantly, test.

Everyone

### Additional Context

_No response_

lvlcn-t:
## Motivation

To simplify the check (config) reconciliation logic. The previous implementation used a map of check names to checks, which added unnecessary complexity to the registration, deletion, and updating of checks. 

Closes #98 

## Changes

A new struct, `ChecksController`, has been introduced to handle the reconciliation logic of the Checks. This component can start, act, and shutdown autonomously, making the registration, update, and removal of checks simpler and more straightforward. 

In addition, a `runtime.Checks` struct has been introduced to hold the checks in a slice and provide thread-safe methods to add, delete, and iterate over the checks. This struct complements the `runtime.Config` struct, which holds the configuration for the checks. The checks are now directly held in a dynamic slice, further simplifying the logic and making the code easier to understand and maintain.

The changes include:

- Creation of the `ChecksController` and `runtime.Checks` structs.
- Refactoring of the reconciliation logic to use the new `ChecksController` and `runtime.Checks`.
- Updates to the functions and methods that previously interacted with the map of checks to now interact with the `ChecksController` and `runtime.Checks`.

## Tests done

I've provided several new tests.

- [x] Unit tests succeeded.
- [x] E2E tests succeeded.

### Manual e2e tests

Logs:
```
$ go run main.go run --config .tmp/config/start-config.yaml 
Using config file: .tmp/config/start-config.yaml
{"time":"2024-02-08T12:57:08.155606092+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/cmd.NewCmdRun.run.func1","file":"/home/installadm/dev/github/sparrow/cmd/run.go","line":81},"msg":"Running sparrow"}
{"time":"2024-02-08T12:57:08.15570043+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/sparrow/targets.(*gitlabTargetManager).Reconcile","file":"/home/installadm/dev/github/sparrow/pkg/sparrow/targets/gitlab.go","line":81},"msg":"Starting global gitlabTargetManager reconciler"}
{"time":"2024-02-08T12:57:08.15594968+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/api.(*api).Run.func1","file":"/home/installadm/dev/github/sparrow/pkg/api/api.go","line":76},"msg":"Serving Api","addr":":8080"}
{"time":"2024-02-08T12:57:38.179964595+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*FileLoader).Run","file":"/home/installadm/dev/github/sparrow/pkg/config/file.go","line":79},"msg":"Successfully got local runtime configuration"}
{"time":"2024-02-08T12:57:38.180136378+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/checks/dns.(*DNS).Run","file":"/home/installadm/dev/github/sparrow/pkg/checks/dns/dns.go","line":99},"msg":"Starting dns check","interval":"20s"}
{"time":"2024-02-08T12:57:38.180178287+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/checks/health.(*Health).Run","file":"/home/installadm/dev/github/sparrow/pkg/checks/health/health.go","line":91},"msg":"Starting healthcheck","interval":"10s"}
{"time":"2024-02-08T12:57:38.18027275+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/checks/latency.(*Latency).Run","file":"/home/installadm/dev/github/sparrow/pkg/checks/latency/latency.go","line":97},"msg":"Starting latency check","interval":"20s"}
{"time":"2024-02-08T12:58:08.181125735+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*FileLoader).Run","file":"/home/installadm/dev/github/sparrow/pkg/config/file.go","line":79},"msg":"Successfully got local runtime configuration"}
{"time":"2024-02-08T12:58:08.392343533+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/sparrow/gitlab.(*Client).FetchFiles","file":"/home/installadm/dev/github/sparrow/pkg/sparrow/gitlab/gitlab.go","line":149},"msg":"Successfully fetched all target files","files":2}
{"time":"2024-02-08T12:58:38.182087954+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*FileLoader).Run","file":"/home/installadm/dev/github/sparrow/pkg/config/file.go","line":79},"msg":"Successfully got local runtime configuration"}
{"time":"2024-02-08T12:59:08.182573733+01:00","level":"INFO","source":{"function":"github.com/caas-team/sparrow/pkg/config.(*FileLoader).Run","file":"/home/installadm/dev/github/sparrow/pkg/config/file.go","line":79},"msg":"Successfully got local runtime configuration"}
```
First reconcilation interval:
```
# HELP sparrow_dns_check_count Total number of DNS checks performed on the target and if they were successful.
# TYPE sparrow_dns_check_count counter
sparrow_dns_check_count{target="10.x.x.x"} 5
sparrow_dns_check_count{target="www.t-systems.com"} 5
sparrow_dns_check_count{target="www.telekom.de"} 5
# HELP sparrow_dns_duration Histogram of response times for DNS checks in seconds.
# TYPE sparrow_dns_duration histogram
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.005"} 4
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.01"} 4
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.025"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.05"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.1"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.25"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="0.5"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="1"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="2.5"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="5"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="10"} 5
sparrow_dns_duration_bucket{target="10.x.x.x",le="+Inf"} 5
sparrow_dns_duration_sum{target="10.x.x.x"} 0.018343466000000003
sparrow_dns_duration_count{target="10.x.x.x"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.005"} 4
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.01"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.025"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.05"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.1"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.25"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="0.5"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="1"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="2.5"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="5"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="10"} 5
sparrow_dns_duration_bucket{target="www.t-systems.com",le="+Inf"} 5
sparrow_dns_duration_sum{target="www.t-systems.com"} 0.018483263
sparrow_dns_duration_count{target="www.t-systems.com"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.005"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.01"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.025"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.05"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.1"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.25"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="0.5"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="1"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="2.5"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="5"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="10"} 5
sparrow_dns_duration_bucket{target="www.telekom.de",le="+Inf"} 5
sparrow_dns_duration_sum{target="www.telekom.de"} 0.016302045
sparrow_dns_duration_count{target="www.telekom.de"} 5
# HELP sparrow_dns_duration_seconds Duration of DNS resolution attempts in seconds.
# TYPE sparrow_dns_duration_seconds gauge
sparrow_dns_duration_seconds{target="10.x.x.x"} 0.000708375
sparrow_dns_duration_seconds{target="www.t-systems.com"} 0.003439707
sparrow_dns_duration_seconds{target="www.telekom.de"} 0.003415548
# HELP sparrow_dns_status Specifies if the target can be resolved.
# TYPE sparrow_dns_status gauge
sparrow_dns_status{target="10.x.x.x"} 1
sparrow_dns_status{target="www.t-systems.com"} 1
sparrow_dns_status{target="www.telekom.de"} 1
# HELP sparrow_health_up Health of targets
# TYPE sparrow_health_up gauge
sparrow_health_up{target="https://gitlab.devops.telekom.de"} 1
sparrow_health_up{target="https://www.example.com"} 1
sparrow_health_up{target="https://www.google.com"} 1
sparrow_health_up{target="https://www.telekom.de"} 1
# HELP sparrow_latency_count Count of latency checks done
# TYPE sparrow_latency_count counter
sparrow_latency_count{target="https://example.com"} 5
sparrow_latency_count{target="https://google.com"} 5
# HELP sparrow_latency_duration Latency of targets in seconds
# TYPE sparrow_latency_duration histogram
sparrow_latency_duration_bucket{target="https://example.com",le="0.005"} 0
sparrow_latency_duration_bucket{target="https://example.com",le="0.01"} 0
sparrow_latency_duration_bucket{target="https://example.com",le="0.025"} 0
sparrow_latency_duration_bucket{target="https://example.com",le="0.05"} 0
sparrow_latency_duration_bucket{target="https://example.com",le="0.1"} 3
sparrow_latency_duration_bucket{target="https://example.com",le="0.25"} 4
sparrow_latency_duration_bucket{target="https://example.com",le="0.5"} 5
sparrow_latency_duration_bucket{target="https://example.com",le="1"} 5
sparrow_latency_duration_bucket{target="https://example.com",le="2.5"} 5
sparrow_latency_duration_bucket{target="https://example.com",le="5"} 5
sparrow_latency_duration_bucket{target="https://example.com",le="10"} 5
sparrow_latency_duration_bucket{target="https://example.com",le="+Inf"} 5
sparrow_latency_duration_sum{target="https://example.com"} 0.800343934
sparrow_latency_duration_count{target="https://example.com"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="0.005"} 0
sparrow_latency_duration_bucket{target="https://google.com",le="0.01"} 0
sparrow_latency_duration_bucket{target="https://google.com",le="0.025"} 0
sparrow_latency_duration_bucket{target="https://google.com",le="0.05"} 0
sparrow_latency_duration_bucket{target="https://google.com",le="0.1"} 3
sparrow_latency_duration_bucket{target="https://google.com",le="0.25"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="0.5"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="1"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="2.5"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="5"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="10"} 5
sparrow_latency_duration_bucket{target="https://google.com",le="+Inf"} 5
sparrow_latency_duration_sum{target="https://google.com"} 0.544606929
sparrow_latency_duration_count{target="https://google.com"} 5
# HELP sparrow_latency_duration_seconds Latency with status information of targets
# TYPE sparrow_latency_duration_seconds gauge
sparrow_latency_duration_seconds{status="200",target="https://example.com"} 0.100032163
sparrow_latency_duration_seconds{status="200",target="https://google.com"} 0.100295391
```
Second reconcilation interval:
```
# HELP sparrow_health_up Health of targets
# TYPE sparrow_health_up gauge
sparrow_health_up{target="https://gitlab.devops.telekom.de"} 1
sparrow_health_up{target="https://www.example.com"} 1
sparrow_health_up{target="https://www.google.com"} 1
sparrow_health_up{target="https://www.telekom.de"} 1
```

## TODO

- [X] I've assigned this PR to myself
- [X] I've labeled this PR correctly


pkg/sparrow/run.go
```diff
@@ -49,16 +49,15 @@ type Sparrow struct {
 	checkCoordinator
 }
 
+// checkCoordinator is used to coordinate the checks and the reconciler
 type checkCoordinator struct {
```
puffitos, line 53: do we really still need this struct? Can't the Controller replace it completely. It seems like an unneeded indirection.

pkg/sparrow/run.go
```diff
@@ -135,49 +136,46 @@ func (s *Sparrow) Run(ctx context.Context) error {
 	}
 }
 
-// ReconcileChecks registers new Checks, unregisters removed Checks,
-// resets the Configs of Checks and starts running the checks
+// ReconcileChecks reconciles the checks.
+// It registers new checks, updates existing checks and unregisters checks not in the new config.
 func (s *Sparrow) ReconcileChecks(ctx context.Context, cfg runtime.Config) {
```
puffitos, line 141: I'm angry at myself, because this function should be a part of the Controller you wrote. And it currently can't be because of the targetManager.

pkg/sparrow/run.go
```diff
@@ -290,14 +223,16 @@ func (s *Sparrow) shutdown(ctx context.Context) {
...
+			sErrs.errTarMan = s.tarMan.Shutdown(ctx)
 		}
-		errA := s.api.Shutdown(ctx)
+		sErrs.errAPI = s.api.Shutdown(ctx)
 		s.loader.Shutdown(ctx)
-		if errS != nil || errA != nil {
-			log.Error("Failed to shutdown gracefully", "contextError", errC, "apiError", errA, "targetError", errS)
+		sErrs.errCheckCont = s.controller.Shutdown(ctx)
+
+		if sErrs.errTarMan != nil || sErrs.errAPI != nil || sErrs.errCheckCont != nil {
```
puffitos, line 234: A sErrs.HasErr() or something similar seems appropriate.

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,226 @@
...
+		db:      dbase,
+		metrics: metrics,
+		checks:  runtime.Checks{},
+		cErr:    make(chan error, 1),
+		done:    make(chan struct{}, 1),
+	}
+}
+
+// ListenErrors listens for errors and unregisters failed checks.
+func (cc *ChecksController) ListenErrors(ctx context.Context) {
```
puffitos, line 55: HandleErrors feels like a more appropriate name. We're not listening to a socket or a connection of some kind. We just handling errors as they come in through the channel.

puffitos (changes requested):
Just a few remarks here and there. I still have to look at the logic of the checkReconciliation and evaluate the whole refactoring.

pkg/sparrow/run.go
lvlcn-t: We can if we move the target enrichment out of the `ReconcileChecks` method.

puffitos: I think that the check controller should have a Run method as well, as do all other components. It should just react to the new `runtime.Config` instances coming in via the channel and register/update/delete checks from the state.

I'm excited to see how you'll separate the enrichTargets from the sparrow.

puffitos: I was taking the channel route again (a channel for the `globalTargets` read from the `targetManager`), as one way of decoupling the `targetManager` from the sparrow and making the targets available to those who need it.

lvlcn-t: I'd rather not implement the `Run` method as long as your target manager channel suggestion isn't implemented.

A `Run` method currently would complicate the `CheckController` with adding an unnecessary channel, because both, the sparrow, and the controller would need the `cRuntime` channel. The sparrow would then send the enriched config through the channel to the controller. 

But if we'll refactor the behavior of the target manager, the `Run` method should be implemented.

lvlcn-t:
Additionally this PR should also close #36 and #43

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+
+	// Unregister checks not in the new config
+	for _, c := range unregList {
+		err = cc.UnregisterCheck(ctx, c)
+		if err != nil {
+			log.ErrorContext(ctx, "Failed to unregister check", "check", c.Name(), "error", err)
+		}
+	}
```
NiklasTreml, line 127: Can you elaborate why this unreglist is created, instead of directly Unregistering in the Update loop? Is there a technical reason or is this just for simplicity?

NiklasTreml (approved):
One minor thing, otherwise LGTM

pkg/sparrow/controller.go
lvlcn-t: The `unregList` is to avoid modifying the `cc.checks` slice while iterating over it. This can cause issues because it's generally not safe to modify a slice while iterating over it because it could lead to a runtime panic. Here a simple example of it: https://go.dev/play/p/uzLeQrqb0_A?v=goprev.go

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+
+	go func() {
+		err := cc.runCheck(ctx, check)
+		if err != nil {
+			log.ErrorContext(ctx, "Failed to start check", "error", err)
+			cc.cErr <- err
+		}
+	}()
+
```
puffitos, line 156: This function, which is called RegisterCheck, does multiple things:

- It registers the check in the metrics registry (that's fine)
- It adds the check to the state (that's exactly what the name says)
- It starts running the check.

The latter is deeply problematic. To run the check, we start a goroutine in this function (the function will return in a bit but the goroutine will keep on going on). The goroutine calls another function (`runCheck`), which starts a separate goroutine, which finally (!!!) initiates the `Run` method of the `Check`. It then has a `for` to catch the results coming in. 

Still, we're not done. The `Run` method kinda does the same. It just has a `for` to catch the results of the internal implementation of how the check must run, which (suprise!) is a function which spawns multiple goroutines (which is fine IMO).

We have to watch out for multiple indirections. Too much abstraction will break our code. We should rethink how the checks start running, as this seems too convoluted, in my humble opinion. 

I think this can be addressed together with refactoring the `CheckController` to start with `Run` method and just do its thing (sorry, I sound like a broken record ATM :D ).

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+		return err
+	}
+
+	for i, existingCheck := range cc.checks.Iter() {
+		if existingCheck.Name() == check.Name() {
+			cc.checks.Delete(i)
+			break
+		}
+	}
+
```
puffitos, line 184: I think it would make sense to refactor this to a `removeCheck(c check)` for the controller. We should try and hide away how the state is implemented (the current implementation is a slice of `Checks` but we shouldn't care about it). In case we want to switch implementations again (choose a map), such functions guard us from having to change too much of our codebase.

I strongly think this isn't a YAGNI feature but a needed part of the behavior of the ChecksController. As the `RegisterCheck` and `UnregisterCheck` dictate the controller's business logic, similar functions should handle its internal state (like the proposed `removeCheck`).

pkg/sparrow/controller.go
puffitos: I was like: "the check only runs once when it's registered and then it's over?". And then I was like "Oh no, the checks only start running when a new config comes in". And then I was like "The checks start running forever when registred but when a new check comes in, they are also updated".

I think we can find a better way to handle this :)

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+	return &ChecksController{
+		db:      dbase,
+		metrics: metrics,
+		checks:  runtime.Checks{},
+		cErr:    make(chan error, 1),
+		done:    make(chan struct{}, 1),
+	}
+}
+
+// HandleErrors listens for errors and unregisters failed checks.
```
puffitos, line 55: HandleErrors doesn't listen, it just handles incoming errors during the controller's lifecycle. :) Pedantic Bruno is pedantic!

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+			err = c.SetConfig(conf)
+			if err != nil {
+				log.ErrorContext(ctx, "Failed to set config for check", "check", c.Name(), "error", err)
+			}
+			delete(newChecks, c.Name())
+		} else {
+			unregList = append(unregList, c)
+		}
+	}
+
```
puffitos, line 120: My hate for if-else knows no end. Please consider the following rewrite:

```go
	for _, c := range cc.checks.Iter() {
		conf := cfg.For(c.Name())
		if conf == nil {
			unregList = append(unregList, c)
			continue
		}
		err = c.SetConfig(conf)
		if err != nil {
			log.ErrorContext(ctx, "Failed to set config for check", "check", c.Name(), "error", err)
		}
		delete(newChecks, c.Name())
	}
```

puffitos (commented):
The refactoring LGTM. Just a few more suggestions and a bigger discussion point which we should turn into an issue.

pkg/sparrow/controller.go
```diff
@@ -0,0 +1,270 @@
...
+		case err := <-cc.cErr:
+			var runErr *ErrRunningCheck
+			if errors.As(err, &runErr) {
+				uErr := cc.UnregisterCheck(ctx, runErr.Check)
+				if uErr != nil {
+					log.ErrorContext(ctx, "Failed to unregister check", "check", runErr.Check.Name(), "error", uErr)
+				}
+			}
+			log.ErrorContext(ctx, "Error while running check", "error", err)
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
lvlcn-t: I'm now returning the error here back to the sparrow that then triggers the graceful shutdown of everything.

pkg/sparrow/controller.go
lvlcn-t: I could also partly refactor it now to call `cc.runCheck` and then there the two go routines are triggered:
```go
func (cc *ChecksController) runCheck(ctx context.Context, check checks.Check) {
	go func() {
		err := check.Run(ctx)
		if err != nil {
			log.ErrorContext(ctx, "Failed to run check", "error", err)
			cc.cErr <- &ErrRunningCheck{
				Check: check,
				Err:   err,
			}
		}
	}()

	go func() {
		err := cc.fanInResults(ctx, check)
		if err != nil {
			log.ErrorContext(ctx, "Failed to fan in results", "error", err)
			cc.cErr <- err
		}
	}()
}
```
That way we can avoid creating a go routine in a go routine (at least in the controller).

lvlcn-t: @puffitos and I had a call regarding this. We came to the following conclusion:

We'll remove the `ResultChan` function from the Checks interface again and reimplement a `fanInChecks` functionality so the controller initializes one Result Channel on its initialization and sets exactly this result channel for each check.

lvlcn-t: I've refactored it in another branch and opened a PR to this branch to address the discussion: #106

lvlcn-t:
I'll resolve merge conflicts after #106 is merged into this branch.

pkg/checks/runtime/checks.go
```diff
@@ -0,0 +1,52 @@
...
+	defer c.mu.Unlock()
+	c.checks = append(c.checks, check)
+}
+
+// Delete deletes a check.
+func (c *Checks) Delete(index int) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.checks = append(c.checks[:index], c.checks[index+1:]...)
+}
```
puffitos, line 45: The `Delete` should do the oposite of the `Add` function. If the `Add` function adds a check to the state, the `Delete` function should also remove a `Check` from the state. 

Passing an integer index defeats the purpose of encapsulation. The function should just Delete a `Check` from its internal state, regardless of whether it's a list or a hashmap.

puffitos (changes requested):
Just a remark on encapsulation of the `Checks` struct.

puffitos (approved):

Issue #101, opened 2024-02-07
Feature: make the update feature of the target manager configurable


puffitos:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

The update functionality of the target manager (update with current timestamp) should be configurable. This allows the target manager to activate/deactivate the update registration feature.

### Solution Description

Make the update-registration feature configurable (enabled/disable) and separate the interval from the registration interval. The two features should be separate.

PR #103, opened 2024-02-08
feat/registration-update

puffitos:
## Motivation

Part of https://github.com/caas-team/sparrow/issues/101

I'll split this into two PRs to keep them easy to merge. First comes the cleanup and refactoring needed and then comes the feature itself. 

## Changes

I've added a more general `Config` struct to the `targetManager`, which includes the general config each `targetManager` should have. This will also include a new field in a coming PR, to regulate the update interval.

I've added a simple Validation function for the general config, to prerpare for the new feature (0 duration will disable the update & registration). A zero checkInterval makes no sense, so it's not allowed.

The rest ist fixing tests and moving the configs to the `targets` package, as we've done with other PRs.

## Tests done

- [x] New unittests

## Future plans

We can regulate the update & registration interval by using the magic number 0 (which is a possible value for the interval). Since the practice to use that number is common, we should adopt it, instead of just dropping the value as not-OK.

## TODO

### DIFF
--- a/pkg/sparrow/targets/errors.go
+++ b/pkg/sparrow/targets/errors.go
@@ -1,3 +1,21 @@
+// sparrow
+// (C) 2023, Deutsche Telekom IT GmbH
+//
+// Deutsche Telekom IT GmbH and all other contributors /
+// copyright owners license this file to you under the Apache
+// License, Version 2.0 (the "License"); you may not use this
+// file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package targets
 
 import "errors"
--- a/pkg/sparrow/targets/errors.go
+++ b/pkg/sparrow/targets/errors.go
@@ -1,0 +1,12 @@
+package targets
+
+import "errors"
+
+// ErrInvalidCheckInterval is returned when the check interval is invalid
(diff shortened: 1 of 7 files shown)


pkg/sparrow/targets/targetmanager_test.go
```diff
@@ -0,0 +1,60 @@
+package targets
```
lvlcn-t, line 1: add license header

pkg/sparrow/targets/errors.go
```diff
@@ -0,0 +1,12 @@
+package targets
```
lvlcn-t, line 1: add license header

lvlcn-t (approved):
you can add the license headers also in #104, so LGTM 😄

pkg/sparrow/targets/gitlab.go
```diff
@@ -80,8 +76,8 @@ func (t *gitlabTargetManager) Reconcile(ctx context.Context) error {
 	log := logger.FromContext(ctx)
 	log.Info("Starting global gitlabTargetManager reconciler")
 
-	checkTimer := time.NewTimer(t.checkInterval)
-	registrationTimer := time.NewTimer(t.registrationInterval)
+	checkTimer := time.NewTimer(t.cfg.CheckInterval)
+	registrationTimer := time.NewTimer(t.cfg.RegistrationInterval)
```
NiklasTreml, line 80: Can you add a unit test for the 0 case, where the update should never happen? I'm not 100% sure but I think a zero timer returns at least immediately, which would cause the opposite of what we want, 1000s of refreshes every second

NiklasTreml (changes requested):

pkg/sparrow/targets/gitlab.go
NiklasTreml: looked into https://github.com/golang/go/issues/10124, you've already implemented there, very good 👍

NiklasTreml (approved):

lvlcn-t (approved):

PR #106, opened 2024-02-13
refactor: use one result channel for all checks

lvlcn-t:
## Motivation

Addresses this discussion: https://github.com/caas-team/sparrow/pull/102#discussion_r1484805340

## Changes

I've removed the result channels of each check. Instead I'm giving the result channel to the `Run` method of each check.

## Tests done

Unit tests succeeded.

## TODO

### DIFF
--- a/pkg/checks/dns/dns_test.go
+++ b/pkg/checks/dns/dns_test.go
@@ -193,9 +193,6 @@ func TestDNS_Run(t *testing.T) {
 			}()
 
 			r := <-cResult
-			if r.Result.Err != tt.want.Err {
-				t.Errorf("DNS.Run() = %v, want %v", r.Result.Err, tt.want.Err)
-			}
 			assert.IsType(t, tt.want.Data, r.Result.Data)
 
 			got := r.Result.Data.(map[string]result)
--- a/pkg/checks/dns/dns_test.go
+++ b/pkg/checks/dns/dns_test.go
@@ -53,7 +53,6 @@ func TestDNS_Run(t *testing.T) {
 				return &DNS{
 					CheckBase: checks.CheckBase{
 						Mu:       sync.Mutex{},
-						ResChan:  make(chan checks.Result, 1),
 						DoneChan: make(chan struct{}, 1),
 					},
 				}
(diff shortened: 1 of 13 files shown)


pkg/checks/dns/dns_test.go
```diff
@@ -190,13 +192,13 @@ func TestDNS_Run(t *testing.T) {
 				}
 			}()
 
-			r := <-c.ResultChan()
-			if r.Err != tt.want.Err {
-				t.Errorf("DNS.Run() = %v, want %v", r.Err, tt.want.Err)
+			r := <-cResult
+			if r.Result.Err != tt.want.Err {
+				t.Errorf("DNS.Run() = %v, want %v", r.Result.Err, tt.want.Err)
 			}
```
puffitos, line 198: We have no case where an error is present in the result struct. Is this an omission on our part or does the `Result` not a good match for the DNS check? The Result.err is always empty in the `Run` function of the DNS check...

pkg/sparrow/controller.go
```diff
@@ -47,17 +48,20 @@ func NewChecksController(dbase db.DB, metrics Metrics) *ChecksController {
...
+		cResult: make(chan checks.ResultDTO, 8), //nolint:gomnd // Buffered channel to avoid blocking the checks
 		cErr:    make(chan error, 1),
 		done:    make(chan struct{}, 1),
 	}
 }
 
-// HandleErrors handles errors that occur while running checks.
-func (cc *ChecksController) HandleErrors(ctx context.Context) error {
+// Run runs the ChecksController with handling results and errors.
+func (cc *ChecksController) Run(ctx context.Context) error {
```
puffitos, line 58: Finally, we have a nice and simple Run method. Well done.

puffitos (commented):
LGTM. There is this `Err` field in the Result struct, which doesn't appear to be used anywhere. We should probably take a look at that.

pkg/checks/dns/dns_test.go
puffitos: Wanna remove it now? I don't feel like we should open an issue for this.

lvlcn-t: Sure, I've removed it from the struct and all tests. 😄

PR #108, opened 2024-02-16
feat: disable continuous loading of check config

### LINKED ISSUE #107: Bug: Panic if loader interval is not set

### Problem Description

Currently the loaders would panic if no interval is set because it then calls [`time.NewTicker(0)`](https://go.dev/src/time/tick.go) which panics.

### Solution Description

We should at least validate if the interval is > 0.

But imo this offers the opportunity to introduce another feature to the loaders.

We could disable the continuous loading of the check config with an interval of 0. This would also align with the feature introduced by #101 which kinda does this for the instance registration and update.

If we introduced this proposal, we'd only need to check our startup config for interval < 0.

Everyone

### Additional Context

_No response_

lvlcn-t:
## Motivation

Closes #107 

## Changes

* fix: validate loader interval
* feat: get runtime cfg once if interval is zero
* test: add test cases for the new interval
* fix: file loader run test race condition

## Tests done

Provided a new unit test for both loaders.

- [x] Unit tests succeeded.
- [x] E2E tests succeeded.

## TODO

### DIFF
--- a/pkg/config/file.go
+++ b/pkg/config/file.go
@@ -58,6 +58,18 @@ func (f *FileLoader) Run(ctx context.Context) error {
 	ctx, cancel := logger.NewContextWithLogger(ctx)
 	defer cancel()
 	log := logger.FromContext(ctx)
+
+	if f.config.Interval == 0 {
+		cfg, err := f.getRuntimeConfig(ctx)
+		if err != nil {
+			log.Warn("Could not get local runtime configuration", "error", err)
+		}
+
+		f.cRuntime <- cfg
+		log.Info("File Loader disabled")
+		return nil
+	}
+
 	tick := time.NewTicker(f.config.Interval)
 	defer tick.Stop()
 
--- a/pkg/config/file.go
+++ b/pkg/config/file.go
@@ -53,7 +53,6 @@ func NewFileLoader(cfg *Config, cRuntime chan<- runtime.Config) *FileLoader {
 
 // Run gets the runtime configuration from the local file.
 // The config will be loaded periodically defined by the loader interval configuration.
-// Returns an error if the loader is shutdown or the context is done.
 func (f *FileLoader) Run(ctx context.Context) error {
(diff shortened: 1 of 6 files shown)


pkg/config/file.go
```diff
@@ -58,6 +58,18 @@ func (f *FileLoader) Run(ctx context.Context) error {
 	ctx, cancel := logger.NewContextWithLogger(ctx)
 	defer cancel()
 	log := logger.FromContext(ctx)
+
+	if f.config.Interval == 0 {
+		cfg, err := f.getRuntimeConfig(ctx)
+		if err != nil {
+			log.Warn("Could not get local runtime configuration", "error", err)
+		}
```
y-eight, line 66: You should return here as well since cfg is nil.

pkg/config/http.go
```diff
@@ -56,7 +56,24 @@ func (hl *HttpLoader) Run(ctx context.Context) error {
...
+	getConfigRetry := helper.Retry(func(ctx context.Context) (err error) {
+		runtimeCfg, err = hl.getRuntimeConfig(ctx)
+		return err
+	}, hl.cfg.Http.RetryCfg)
+
+	if hl.cfg.Interval == 0 {
+		err := getConfigRetry(ctx)
+		if err != nil {
+			log.Warn("Could not get local runtime configuration", "error", err)
+		}
```
y-eight, line 70: same here

pkg/config/validate.go
```diff
@@ -40,6 +40,11 @@ func (c *Config) Validate(ctx context.Context) error {
 		log.Error("The name of the sparrow must be DNS compliant")
 	}
 
+	if c.Loader.Interval < 0 {
+		ok = false
+		log.Error("The loader interval should be above 0", "interval", c.Loader.Interval)
```
y-eight, line 45: or equal

y-eight (approved):
LGTM

puffitos (approved):
LGTM

Issue #112, opened 2024-02-21
Feature: Use tcp for traceroute


niklastreml:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

The current implementation of the traceroute check uses udp to perform its logic. This is not ideal since udp is not connection oriented. This means that the check can never really know whether a packet reached its destination, unless that destination sends back an ICMP packet, which tends to be blocked by firewalls.

### Solution Description

We can get around the above issues by reimplementing the traceroute functionality on top of TCP instead of UDP. We can essentially send a TCP SYN packet to initiate a handshake with a server on its open port. The underlying logic of the check stays the same, we're only changing the protocol.

### Who can address the issue?

Anyone who wants to read some RFCs and feels like playing around with berkeley sockets

### Additional Context

https://en.wikipedia.org/wiki/Berkeley_sockets
https://www.ietf.org/rfc/rfc793.txt#:~:text=Transmission%20Control%20Protocol-,3.%20%20FUNCTIONAL%20SPECIFICATION,-3.1.%20%20Header%20Format
https://github.com/mct/tcptraceroute/blob/master/probe.c#L80

lvlcn-t:
I've started to implement this in [feat/tcp-traceroute](https://github.com/caas-team/sparrow/tree/feat/tcp-traceroute). Unfortunately, we cannot get any useful information with a pure tcp traceroute because of the nature of the tcp protocol. We can only know how many hops we have done without a successful connection (a response in form of a RST packet) but any further information is not possible because the routers in between only send back an ICMP packet if the ttl has expired (traceroute uses this to get the hops). I also implemented the other forms of traceroute in the branch (udp/icmp & raw sockets/icmp) to allow the user to choose the method they want to use but as you may know both of these methods either need the `CAP_NET_RAW` capability or root privileges which is not ideal. I will continue to work on this but I wanted to share my progress so far. Additionally I'm planning to adjust the user configuration options as follows:

```yaml
traceroute:
  protocol: tcp
  interval: 10s
  timeout: 400s
  maxHops: 64
  retry:
    count: 5
    delay: 1s
  targets:
    - addr: example.com
      port: 80
    - addr: google.com
      port: 80
```

This way the user can choose the protocol they want to use. I'm also planning to adjust the API response to the following schema:

```go
// result represents the result of a single hop in the traceroute
type result struct {
	// Duration represents the total duration of the traceroute
	Duration float64
	// Hops represents the hops to the target
[...]

Issue #111, opened 2024-02-21
Feature: OpenTelemtry


NiklasTreml:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

The prometheus format is not sufficient for exporting metrics from the traceroute check, aside from the amount of hops. To provide more complex data, it might be necessary to support a second data format like opentelemetry.

### Solution Description

Implement the [OTEL library](https://github.com/open-telemetry/opentelemetry-go) and inject it into the checks, so they can write their own traces. Traceroute can use this to collect more detailed data about how every single invocation of the check behaves, essentially allowing a user to visualize how packets move from sparrow to their target

### Additional Context

https://github.com/open-telemetry/opentelemetry-go
https://opentelemetry.io/docs/languages/go/
https://www.jaegertracing.io/docs/1.54/client-libraries/#go
https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/traces/

NiklasTreml:
There might be a way to build traceroute metrics with prometheus, but I'm not sure if I like it.
We can create a sparrow_traceroute_hop metric that looks sort of like this:
```
sparrow_traceroute_hop{target="somedomain.com" n=2 hopAddr="10.0.0.1"} 152
```

- **target** is what the user wants to check in the config
- **hopAddr** is the specific address where the packet failed
- **n** declares that **hopAddr** is the **n**th hop on the way to **target**
- the actual metric is the time in milliseconds it took to get to that hop and back

The following traceroute cli output would map to prometheus metrics like so:
```bash
$ traceroute 8.8.8.8
traceroute to 8.8.8.8 (8.8.8.8), 30 hops max, 60 byte packets
 1  MyRouter (68.87.192.226)         0.345 ms
 2  68.87.192.227 (68.87.192.227)    9.291 ms
 3  dns.google (8.8.8.8)             9.231 ms
```

```
sparrow_traceroute_hop{target="8.8.8.8" n=1 hopAddr="68.87.192.226"} 0.345
sparrow_traceroute_hop{target="8.8.8.8" n=2 hopAddr="68.87.192.227"} 9.291
sparrow_traceroute_hop{target="8.8.8.8" n=3 hopAddr="8.8.8.8"} 9.231
```

I'm not sure how easy it is to create a nice visualization from this though.

y-eight:
I think it is not a good idea. Especially when the route is changing often. The metrics do not provide a dependency to each other over time.

[...]

PR #113, opened 2024-02-21
Feat/traceroute

niklastreml:
## Motivation
This pr implements a basic traceroute check. 

Relates to #111, #112

Currently its using "github.com/aeden/traceroute" to implement the actual tracerouting logic, but I intend to to change this in the near future (more on this in #112). 

## In case you're testing this locally
The check is extremely slow, since it creates a lot of network requests. I tried to improve by concurrently running the check for every target, but the internals of the library do everything synchronously so there isn't much we can actually improve there outside of #112. While developing I used the below config, with retries set to 0, and it took a minute or so per run of the check. 
### Config
```yaml
traceroute:
  interval: 5s
  timeout: 3s
  retries: 3
  maxHops: 8
  targets:
    - addr: 8.8.8.8
```

### Debugging
Since this check requires extra permissions to open the required sockets, debugging is a bit complicated. I have added a script in `scripts/debug-elevated.sh` which compiles sparrow as an unoptimized binary and launches it with a headless debugger and elevated permissions. You can then use the `Connect to server` target in the `.vscode/launch.json` to connect to the debugger and debug as usual. This is only necessary, when you actually need to debug the traceroute check, you can debug the others as usual.

## Changes

## TODO


pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+	cResult := make(chan internalResult, len(c.config.Targets))
+
+	for _, t := range c.config.Targets {
+		wg.Add(1)
+		go func(t Target) {
+			defer wg.Done()
+			log.Debug("Running traceroute", "target", t.Addr)
+			start := time.Now()
+			tr, trerr := c.traceroute(t.Addr, int(t.Port), int(c.config.Timeout/time.Millisecond), c.config.Retries, c.config.MaxHops)
+			duration := time.Since(start)
```
lvlcn-t, line 133: This should be in seconds since the other checks also do it in seconds.

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+			if trerr != nil {
+				log.Error("Error running traceroute", "err", trerr, "target", t.Addr)
+			}
+
+			log.Debug("Ran traceroute", "result", tr, "duration", duration)
```
lvlcn-t, line 138: Did you consciously didn't implement the retry mechanism for the check or did you just forget it?

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+	for _, t := range c.config.Targets {
+		wg.Add(1)
+		go func(t Target) {
+			defer wg.Done()
+			log.Debug("Running traceroute", "target", t.Addr)
+			start := time.Now()
+			tr, trerr := c.traceroute(t.Addr, int(t.Port), int(c.config.Timeout/time.Millisecond), c.config.Retries, c.config.MaxHops)
+			duration := time.Since(start)
+			if trerr != nil {
+				log.Error("Error running traceroute", "err", trerr, "target", t.Addr)
```
lvlcn-t, line 135: I think we should append the addr to the logger's attributes.
```go
lo := log.With("target", t.Addr)
```

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+			cResult <- internalResult{addr: t.Addr, res: result}
+		}(t)
+	}
+
+	log.Debug("Waiting for traceroute checks to finish")
+
+	go func() {
+		wg.Wait()
+		close(cResult)
+	}()
```
lvlcn-t, line 161: Is this go routine necessary? I think we can `defer close(cResult)` after initialization.

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+	Addr    string
+	Latency time.Duration
+	Success bool
+}
+
+func (d *Traceroute) Run(ctx context.Context, cResult chan checks.ResultDTO) error {
+	ctx, cancel := logger.NewContextWithLogger(ctx)
+	defer cancel()
+	log := logger.FromContext(ctx)
+	log.Info("Starting dns check", "interval", d.config.Interval.String())
```
lvlcn-t, line 84: Some logs betray you, this is copy pasta! 🍝

pkg/checks/traceroute/traceroute_test.go
```diff
@@ -0,0 +1,149 @@
...
+	type want struct {
+		expected map[string]Result
+	}
+	type testcase struct {
+		name string
+		c    *Traceroute
+		want want
+	}
+
+	cases := []testcase{
```
lvlcn-t, line 26: Please consider refactoring the table driven tests to be aligned with other tests of the repository.

I know many find this approach to table driven tests better but imo it's less readable than just keeping everything in one unnamed struct.

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return checks.ErrConfigMismatch{
+		Expected: CheckName,
+		Current:  cfg.For(),
+	}
+}
+
+// Schema returns an openapi3.SchemaRef of the result type returned by the check
+func (c *Traceroute) Schema() (*openapi3.SchemaRef, error) {
+	return checks.OpenapiFromPerfData[Result](Result{})
```
lvlcn-t, line 202: This should be:
```suggestion
	return checks.OpenapiFromPerfData[map[string]Result](map[string]Result{})
```

lvlcn-t (changes requested):
First review

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+func (c *Config) Validate() error {
+	if c.Timeout <= 0 {
+		return fmt.Errorf("timeout must be greater than 0")
+	}
+	if c.Interval <= 0 {
+		return fmt.Errorf("interval must be greater than 0")
+	}
+	for _, t := range c.Targets {
+		if _, err := url.Parse(t.Addr); err != nil {
+			return fmt.Errorf("%s is not a valid url", t.Addr)
```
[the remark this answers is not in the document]
puffitos: Does it make sense checking for the scheme as well? http/https; or should all all protocols be allowed?

puffitos (commented):
Some remarks:

- please document at least the exposed functions
- Please use a consistent name for the receiver of the `Traceroute` functions (`d` and `c` are used, `tr` maybe be a good candidate)

pkg/checks/traceroute/traceroute.go
puffitos: I think it's because the module already has the retry mechanism implemented ;)

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,227 @@
...
+			DoneChan: make(chan struct{}),
+		},
+	}
+}
+
+type Traceroute struct {
+	checks.CheckBase
+	config     Config
+	traceroute tracerouteFactory
+}
```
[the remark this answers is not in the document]
NiklasTreml: I wanted to keep the pr small, but that's also a factor yes

pkg/checks/traceroute/traceroute.go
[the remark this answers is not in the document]
NiklasTreml: There is no point in checking the scheme as http/Https is a layer 7 concern. We don't really care, the target just has to respond to a UDP packet with an icmp message. This will change a bit when moving to TCP, but the scheme is not a concern and should not be present. The only requirement is a valid IP or DNS name

pkg/checks/traceroute/traceroute.go
NiklasTreml: This is necessary. The loop below reads results from cResult until this channel closed. If we don't close it in some way, the thread would block forever, as no more result are created. By waiting for the waitgroup to finish in a separate goroutine, we can close the channel once we know all targets are done checking. The loop will the run until all results are taken out of the channel and then break

pkg/checks/traceroute/traceroute.go
NiklasTreml: What im doing here ensures that whatever duration is set in the config is converted to milliseconds, since the library requires an integer for the timeout in milliseconds for some reason. I should probably move this into the implementation, i.e. the tracerouteFactory that uses the library.

pkg/checks/traceroute/traceroute.go
NiklasTreml: @puffitos is right. The library implements its own retrying mechanism internally. It would probably make sense to refactor this to take our RetryConfig, so everything is uniform

pkg/checks/traceroute/traceroute.go
lvlcn-t: What I meant with this:
```go
duration := time.Since(start).Seconds()
```
You could also remove the duration since it's only used for a debug log.

pkg/checks/traceroute/traceroute_test.go
lvlcn-t: I think you misunderstood me. If we used an anonymous struct literal like this we wouldn't need to type out the struct for every test case:
```go
tests := []struct{
	name string
	c *Traceroute
	want map[string]Result
}{
	{
		name: "Success 5 hops",
		c:    newForTest(success(5), []string{"8.8.8.8"}),
		want: map[string]Result{
			"8.8.8.8": {
				NumHops: 5,
				Hops: []Hop{
					{Addr: "0.0.0.0", Latency: 0 * time.Second, Success: false},
					{Addr: "0.0.0.1", Latency: 1 * time.Second, Success: false},
					{Addr: "0.0.0.2", Latency: 2 * time.Second, Success: false},
					{Addr: "0.0.0.3", Latency: 3 * time.Second, Success: false},
					{Addr: "google-public-dns-a.google.com", Latency: 69 * time.Second, Success: true},
				},
			},
		},
	},
	{
		name: "Traceroute internal error fails silently",
		c:    newForTest(returnError(&net.DNSError{Err: "no such host", Name: "google.com", IsNotFound: true}), []string{"google.com"}),
		want: map[string]Result{
				"google.com": {Hops: []Hop{}},
		},
	},
}
```

With test cases like this we facilitate easier maintenance and readability. It even reduces the amount of boilerplate code (29 vs. 38 in your version). It's also more readable since you don't need the cognitive load of defining a `want` struct which is only an abstraction of defining the field directly. 

Named structs would only reduce boilerplate code if we used these structs also in other test functions.
Refer to this article for more details on table driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

pkg/checks/traceroute/traceroute.go
NiklasTreml: I don't think this matters since `duration` is a `time.Duration` which the logger should marshal to something readable

pkg/checks/traceroute/traceroute_test.go
NiklasTreml: I prefer creating a local type for this, since you can't always omit the type definition, like [here](https://github.com/caas-team/sparrow/blob/main/pkg/api/api_test.go#L42-L54) for example image

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,199 @@
...
+	var wg sync.WaitGroup
+	cResult := make(chan internalResult, len(tr.config.Targets))
+
+	for _, t := range tr.config.Targets {
+		wg.Add(1)
+		go func(t Target) {
+			defer wg.Done()
+			log.Debug("Running traceroute", "target", t.Addr)
+			start := time.Now()
+			tr, trerr := tr.traceroute(t.Addr, int(t.Port), int(tr.config.Timeout/time.Millisecond), tr.config.Retries, tr.config.MaxHops)
```
lvlcn-t, line 120: Variable shadowing of `tr`

pkg/checks/traceroute/traceroute.go
lvlcn-t: Additionally use the key `error` instead of `err` here to keep it consistent to the rest of the error logs.

pkg/checks/traceroute/traceroute.go
lvlcn-t: You can also rename `trerr` to just `err` here.

pkg/checks/traceroute/config.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (c *Config) For() string {
+	return CheckName
+}
+
+func (c *Config) Validate() error {
+	if c.Timeout <= 0 {
+		return checks.ErrInvalidConfig{CheckName: CheckName, Field: "traceroute.timeout", Reason: "must be greater than 0"}
+	}
+	if c.Interval <= 0 {
```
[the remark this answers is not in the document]
puffitos: I don't think we need to set constants for some arbitrary minimum Interval & Timeout values. We should let the user decide what they want to do. If the want to set a 99ms Interval, that's their issue, not ours.

pkg/checks/traceroute/config.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+	"net"
+	"net/url"
+	"time"
+
+	"github.com/caas-team/sparrow/pkg/checks"
+)
+
+type Config struct {
```
puffitos, line 12: Please comment the struct and the fields, as we want to create automatic documentation for the structs down the line ;)

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,199 @@
...
+
+const CheckName = "traceroute"
+
+type Target struct {
+	Addr string `json:"addr" yaml:"addr" mapstructure:"addr"`
+	Port uint16 `json:"port" yaml:"port" mapstructure:"port"`
+}
+
+// Config is the configuration for the traceroute check
+func NewCheck() checks.Check {
```
puffitos, line 25: I think you forgot to remove the godoc here. Please write a godoc for the constructor.

pkg/checks/traceroute/traceroute_test.go
puffitos: I agree with Tom on this one. Defining a type `want` to only hold an `expected` field, which is a simple `map[string]result` seems unnecessary. Omitting types in a `want` or `expected` field in a table driven test doesn't seem like a great idea as well - the test should express clear intention. 

go is a language of boilerplate and `if err!= nil`. It the language we love, it's ok to have to type something more. Copilot takes care of the most of it anyway :P

puffitos: In your example, I would go for wantMethod, wantPath, wantStatus. If this pattern repeats in multiple tests, then I would go for a refactored `want` struct. The `want` struct is IMO an abstraction which desires to cover everything, which is totally impossible, if more that one thing is expected. 

I think it's better to be verbose in tests, if we don't totally sacrifice maintainability, than to opt for abstractions which just reduce some boilerplate. Each test can and should look different, there's no need for some universal pattern (we already have that with table-driven sets and behavior-driven Testing.

pkg/checks/traceroute/config.go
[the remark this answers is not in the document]
niklastreml: I agree with this, the user should be able to set stupid values. There might be a reason we're not think about, where a really low value actually makes sense

lvlcn-t:
@niklastreml Please also document the check in a new section of the README.

README.md
```diff
@@ -435,6 +438,33 @@ dns:
   - Description: Histogram of response times for DNS checks
   - Labelled with `target`
 
+### Check: Traceroute
```
lvlcn-t, line 441: I think we should write something about the capabilities needed for the check to succeed and that the user can still use the sparrow without these capabilities for the other checks.

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+var _ checks.Check = (*Traceroute)(nil)
+
+const CheckName = "traceroute"
+
+type Target struct {
+	Addr string `json:"addr" yaml:"addr" mapstructure:"addr"`
+	Port uint16 `json:"port" yaml:"port" mapstructure:"port"`
+}
```
lvlcn-t, line 22: > Please comment the struct and the fields, as we want to create automatic documentation for the structs down the line ;)

This is part of the config struct so it should also have godocs,

pkg/checks/traceroute/traceroute.go
```diff
@@ -0,0 +1,199 @@
...
+		close(cResult)
+	}()
+
+	log.Debug("All traceroute checks finished")
+
+	for r := range cResult {
+		res[r.addr] = r.res
+	}
+
+	log.Debug("Finished traceroute checks", "result", res)
```
lvlcn-t, line 157: Duplicate debug logs, please remove the first

lvlcn-t (commented):
Only some minor things, I think after that it'll LGTM.

Issue #79, opened 2024-01-19
Bug: Startupprobe causes gap in data collection


NiklasTreml:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If a sparrow pod is restarted, it causes a gap in monitoring. This is due to the healthcheck returning an HTTP 200 to the kubernetes probe. Since Kubernetes thinks the new sparrow pod is ready to accept traffic, prometheus scrape requests are sent to that new pod immediately, before the startup routine is actually finished. 

Screenshot from prometheus below

### Expected Behavior

A restart of the sparrow deployment waits for the new pod to have finished its first reconciliation cycle. Only then will the pod accept traffic. This ensures, that prometheus never scrapes a pod while still in startup.

### Steps To Reproduce

1. Spin up sparrow on kubernetes
2. Restart the deployment
3. HTTP GET /metrics or /openapi
You will notice sparrow returns no metrics for a few seconds after startup. It also does not show any checks in the /openapi endpoint 

### Relevant logs and/or screenshots, environment information, etc.

![image](https://github.com/caas-team/sparrow/assets/27763017/6d0743ef-a543-434b-aa1c-9eab50ac33db)

lvlcn-t:
I've started to implement this in [feat/startup-probe](https://github.com/caas-team/sparrow/tree/feat/startup-probe). But imo we should wait with implementing this after the #72 and #92 are merged because then it'll be easier to register a `/healthz` endpoint that checks if all configured checks are running properly with their configuration.

lvlcn-t:
I've implemented the other half of this issue, regarding the `/healthz` endpoint I mentioned before, in [feat/healthz-endpoint](https://github.com/caas-team/sparrow/tree/feat/healthz-endpoint). Therefore I've implemented yet another sparrow component named `Checker` that requests each API endpoint and checks if it's healthy (returns a StatusOK).

puffitos:
If the problem. Is the gap in metrics (which is isn't a problem we should concern ourselves with but the people accessing the metrics), then the solution is the following :

As long as the check hasn't produced any metrics, don't expose a metrics endpoint for the check.

I guess this would complicate the check registration process, if we implement something like that.

IMO this is currently a non-issue. The metrics exposed by the check are always the last executed check results. If no check has been run, no results can be delivered. If a check runs once every year, then the same result is delivered almost 365 days long.

[...]

PR #115, opened 2024-02-28
feat: load check config on startup

lvlcn-t:
## Motivation

Partly addresses the conversation of #79 to instantly trigger the config load on sparrow startup.

## Changes

I've added the initial startup config load and refactored the http loader to return the runtime config per value instead of reference to avoid sending a nil pointer through the channel.

## Tests done

I've provided some unit tests for the initial config load.

- [x] Unit tests succeeded.
- [ ] E2E tests succeeded. - we have no e2e tests for this.

## TODO


pkg/sparrow/run.go
```diff
@@ -77,6 +80,7 @@ func New(cfg *config.Config) *Sparrow {
 		cErr:       make(chan error, 1),
 		cDone:      make(chan struct{}, 1),
 		shutOnce:   sync.Once{},
+		checker:    healthz.New(cfg.Api),
```
niklastreml, line 83: Maybe this should just pass the port instead of the whole api config object. We should ensure in some way, that sparrow always checks against localhost:<apiport> for its healthz.

niklastreml: To clarify on this: my worry is, that if a user configures some dns name as their api address (whatever the reason for that), we would use that dns name to check the local endpoints. If we were in kubernetes and used the service name for that, the healthz requests would be sent to the service, which would forward it to one of the **already existing** 'old' sparrows. Since they are already alive, the healthcheck would be green, causing kubernetes to flag the new sparrow as healthy, even if it hasn't yet completed it's startup or is faulty

lvlcn-t: I've added a `formatAddress` utility function that handles various api address formats. 

For more information, look at the unit tests I've provided: https://github.com/caas-team/sparrow/commit/f8cef21e23467e92a58ecd7067c811ee0c06e5d1#diff-d0b860d237946173d5459e969e0e7789b1ef1c55856deb36608a9900856bf798R115

pkg/checks/health/health.go
```diff
@@ -75,8 +76,9 @@ func (h *Health) Run(ctx context.Context, cResult chan checks.ResultDTO) error {
 	ctx, cancel := logger.NewContextWithLogger(ctx)
 	defer cancel()
 	log := logger.FromContext(ctx)
-	log.Info("Starting healthcheck", "interval", h.config.Interval.String())
 
+	log.Info("Starting healthcheck", "interval", h.config.Interval.String())
+	h.Running = true
```
lvlcn-t, line 81: The new changes address the concerns from @puffitos in https://github.com/caas-team/sparrow/pull/116#pullrequestreview-1907143614. I've simplified the `/healthz` endpoint by checking if every registered check is running. Therefore I've added a `IsRunning` method to the Checks interface so the internal state is taken into account.

Now the only question is: **When is a check running?** With how I implemented it now, we wouldn't address #79 because there is still a time difference between the check's `(*checks.Check).Run()` call and the first execution of each check's interval. So to give you an example:

Imagine having the health check configured with an interval of 10mins. Then the pod would be marked as ready even though the check hasn't run yet and will not run for another ~9,9min, resulting again in the problem of the issue:

image

To avoid this behavior, we could set the `Running` field of each check true after the first execution of the check but I that would again strongly connect the checks to the sparrow, which we didn't want to do.

puffitos: Let's discuss this in the issue first.

y-eight (approved):
Looks fine for me.

Still it is not ideal but it is a hard decision. 
@niklastreml also mention to give the user multiple options to define a startup & readiness config for the sparrow. At the moment that is overengineereed I think.

PR #119, opened 2024-03-01
fix/manager-timers

puffitos:
## Motivation

The target manager is broken and only runs once.

## Changes

- Reverted changes
- Added new mock test and made old ones more robust
- Fixed two bugs, where register and update would be called indefinetely, because of not checking for registration first. 

## Tests done

Run the code locally, works as expected. Registration, update and getting targets worked as expected.

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO

### DIFF
--- a/pkg/sparrow/targets/gitlab.go
+++ b/pkg/sparrow/targets/gitlab.go
@@ -133,7 +133,7 @@ func (t *gitlabTargetManager) Shutdown(ctx context.Context) error {
 	ctxS, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
 	defer cancel()
 
-	if t.Registered() {
+	if t.registered {
 		f := gitlab.File{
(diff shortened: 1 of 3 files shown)


pkg/sparrow/targets/gitlab.go
```diff
@@ -159,10 +162,15 @@ func (t *gitlabTargetManager) Shutdown(ctx context.Context) error {
 // in the gitlab repository
 func (t *gitlabTargetManager) register(ctx context.Context) error {
 	log := logger.FromContext(ctx)
-	log.Debug("Registering as global target")
-
 	t.mu.Lock()
 	defer t.mu.Unlock()
+
+	if t.registered {
```
lvlcn-t, line 168: Also use the convenience method here:
```go
if t.Registered()
```

Issue #66, opened 2024-01-11
Feature: Support other remote state backends


lvlcn-t:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Problem Description

As of now, our target manager implementation exclusively supports GitLab through its API for registering the sparrow instances. This limitation restricts the versatility and applicability of our tool, especially for users who utilize other Git-based hosting services.

### Proposed Solution

We should refactor the current target manager to leverage either direct Git operations or requesting the GitLab API depending on what the user has configured in his startup config. Additionally this will universally extend our support to any Git-based remote repository.

After implementing this, the configuration options for the target manager will be expanded to include both Git and GitLab, providing users with the flexibility to connect to their preferred Git hosting service. The proposed configuration schema is as follows:

```yaml
targetManager:
  type: git | gitlab
  checkInterval: 1m
  registrationInterval: 1m
  updateInterval: 1m
  unhealthyThreshold: 3m
  git:
    repoUrl: https://my-git-instance.sparrow.com/my-repo.git
    token: gitpat-xxxxxxxxxxxx
  gitlab:
    baseUrl: https://gitlab.com
    token: glpat-xxxxxxxxxxxx
    projectId: 123456
```

### Who Can Address This Issue?

Go devs

### Additional Context

https://github.com/go-git/go-git

lvlcn-t:
I've started to implement this in [refactor/git-target-manager](https://github.com/caas-team/sparrow/tree/refactor/git-target-manager)

### TODO
- [ ] Provide unit tests for `CommitFile` and `DeleteFile` methods
- [ ] Adjust helm chart values with new target manager config

lvlcn-t:
The new config of the target manager after implementing this issue, will look like this:
```yaml
targetManager:
  type: git | gitlab
  checkInterval: 1m
  registrationInterval: 1m
  updateInterval: 1m
  unhealthyThreshold: 3m
  git:
    repoUrl: https://my-git-instance.sparrow.com/my-repo.git
    token: gitpat-xxxxxxxxxxxx
  gitlab:
    baseUrl: https://gitlab.com
    token: glpat-xxxxxxxxxxxx
    projectId: 123456
```

lvlcn-t:
[...]

PR #120, opened 2024-03-03
refactor: startup config validation

lvlcn-t:
## Motivation

Currently the target manager config has a `Validate` method that is never consumed.

This PR is part of #66 which will implement another target manager type and therefore needs more target manager config validation. That's also when I noticed that the target manager config is never validated. 😄 

## Changes

I've refactored the startup config validation so every startup config component has its own `Validate()` method.

* fix: validate target manager config
* refactor: split config validation in smaller methods
* chore: use errors instead of ok booleans

## Tests done

I've adjusted the existing unit tests.

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO

### DIFF
--- a/pkg/config/errors.go
+++ b/pkg/config/errors.go
@@ -1,0 +1,34 @@
+// sparrow
+// (C) 2024, Deutsche Telekom IT GmbH
+//
+// Deutsche Telekom IT GmbH and all other contributors /
+// copyright owners license this file to you under the Apache
+// License, Version 2.0 (the "License"); you may not use this
+// file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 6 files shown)


lvlcn-t:
The actions are failing because the PR comes from my fork.

All checks are succeeding here, where I've implemented the whole issue of #66 (also with the changes of this PR): https://github.com/caas-team/sparrow/tree/feat/git-tarman

puffitos (approved):
LGTM. I don't appreciate the log in the validate function, that's only code we'll forget to update when the logic changes. We should just return a (structured) error. The standardized error text is OK for now.

Thanks for adding the godocs as well and refactoring the tests :)

lvlcn-t:
@puffitos I agree with you, maybe we should change that in the future to returning to a structured error like this:
```go
// ErrInvalidConfig is an error for invalid configuration
type ErrInvalidConfig struct {
	Field string
	Err   error
}

// Error returns the error message
func (e ErrInvalidConfig) Error() string {
	return fmt.Sprintf("invalid config for %q: %v", e.Field, e.Err)
}
```
But since it makes no difference for the user, I'll implement this in another PR in the near future.
(I want to merge this to continue opening PR's for #66 😄 )

puffitos:
@lvlcn-t the struct seems appropriate. The #111 will take priority in this iteration, just FIY.

PR #121, opened 2024-03-05
Remove unused errors

### LINKED ISSUE #37: Code smell: error is never produced in Shutdown

### Current Behavior

Both the `Health` and `Latency` `Checks` never return an error when running the `Shutdown` implementation. We should either: 

- check for error cases and return them (catch eventual panics that could happen in the channels (`h.done <- true`, done has a capacity of 1)
- remove the error from the function alltogether

### Expected Behavior

Functions that return errors, which are ignored or only logged are a code smell. An error means that something has gone wrong and should be handled.

### Steps To Reproduce

No reproduction needed.

_No response_

_No response_

### Anything else?

_No response_

puffitos:
## Motivation

Closes #37 - no need to carry corpses around in the codebase.

## Changes

See commits. Removed unused err values and ctx values as well from all shutdown functions of the checks, and updated the checkController's shutdown as well (it was never returing an err).

## Tests done

- [x] Unit tests succeeded
- [ ] E2E tests succeeded - not running any as we have none

## TODO

### DIFF
--- a/pkg/sparrow/controller.go
+++ b/pkg/sparrow/controller.go
@@ -65,10 +65,7 @@ func (cc *ChecksController) Run(ctx context.Context) error {
 		case err := <-cc.cErr:
 			var runErr *ErrRunningCheck
 			if errors.As(err, &runErr) {
-				uErr := cc.UnregisterCheck(ctx, runErr.Check)
-				if uErr != nil {
-					log.ErrorContext(ctx, "Failed to unregister check", "check", runErr.Check.Name(), "error", uErr)
-				}
+				cc.UnregisterCheck(ctx, runErr.Check)
 			}
 			log.ErrorContext(ctx, "Error while running check", "error", err)
 		case <-ctx.Done():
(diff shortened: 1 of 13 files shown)


pkg/sparrow/controller.go
```diff
@@ -125,7 +119,7 @@ func (cc *ChecksController) Reconcile(ctx context.Context, cfg runtime.Config) {
 
 	// Unregister checks not in the new config
 	for _, c := range unregList {
-		err = cc.UnregisterCheck(ctx, c)
+		cc.UnregisterCheck(ctx, c)
 		if err != nil {
 			log.ErrorContext(ctx, "Failed to unregister check", "check", c.Name(), "error", err)
 		}
```
lvlcn-t, line 125: This can also be removed

lvlcn-t (commented):
One minor thing, otherwise LGTM

pkg/sparrow/controller.go
puffitos: Removed it, thanks for the review. The `RegistrerCheck` also never seems to return an error, it just pumps everything down the error channel... I'll change that in a later PR.

pkg/sparrow/controller.go
```diff
@@ -83,13 +80,10 @@ func (cc *ChecksController) Run(ctx context.Context) error {
 // Shutdown shuts down the ChecksController.
 func (cc *ChecksController) Shutdown(ctx context.Context) (err error) {
```
lvlcn-t, line 81: You may also remove the error return here, either in this PR or in the later one for #123

pkg/sparrow/controller_test.go
```diff
@@ -302,45 +300,13 @@ func TestChecksController_UnregisterCheck(t *testing.T) {
 			check:   health.NewCheck(),
 			wantErr: false,
```
lvlcn-t, line 301: This field can be removed now

pkg/sparrow/controller.go
puffitos: Later

pkg/sparrow/controller_test.go
puffitos: I'll do this in the #123, as the tests need to be refactored to truly test some use cases.

lvlcn-t (approved):

PR #125, opened 2024-03-14
feat: git remote backend interactor

lvlcn-t:
## Motivation

We currently only support gitlab as remote registration backend through its API. This should change to support all git based repositories. For more information on that please refer to the issue.

Closes #66 

## Changes

I've introduced a new remote interactor type, that is able to perform our defined interactor interfaces actions on any git repository. Currently the git interactor is only able to communicate via HTTP, but that could be changed in a future PR if a user needs ssh communication.

As you've already seen with my last two PRs regarding the issue, I've tried to do it as simple as possible so I hope the code is easy to understand. 😅 

For additional information look at the commits.

## Tests done

I've provided unit tests for the interactor. To do that I've done several things like hiding the remote interactions (Clone, Pull, Push) behind an interface so we can mock it and setup a `repomock` package that sets up a new local repository where the local git interactions are performed. I've consciously decided not to hide the whole go-git library behind an interface because it'd not bring us that much more benefit, but if you think we should still be able to test error cases of local git actions then we'd need to create an interface for it.

- [x] Unit tests succeeded
- [ ] E2E tests succeeded - we have none for the target manager

#### Logs
```sh
❯ go run ./main.go run --config .tmp/start-config.yaml
Using config file: .tmp/start-config.yaml
time=20:29:11 level=INFO source=/home/tom/dev/sparrow/cmd/run.go:89 msg="Running sparrow"
time=20:29:11 level=INFO source=/home/tom/dev/sparrow/pkg/sparrow/targets/manager.go:82 msg="Starting target manager reconciliation"
time=20:29:11 level=INFO source=/home/tom/dev/sparrow/pkg/api/api.go:76 msg="Serving Api" addr=:8080
time=20:29:11 level=INFO source=/home/tom/dev/sparrow/pkg/config/file.go:71 msg="File Loader disabled"
time=20:30:11 level=INFO source=/home/tom/dev/sparrow/pkg/sparrow/targets/remote/git/git.go:82 msg="Successfully fetched all target files" files=0
time=20:30:11 level=DEBUG source=/home/tom/dev/sparrow/pkg/sparrow/targets/manager.go:251 msg="Updated global targets" targets=0
time=20:30:11 level=DEBUG source=/home/tom/dev/sparrow/pkg/sparrow/targets/manager.go:196 msg="Not registered as global target, no update done."
time=20:30:11 level=DEBUG source=/home/tom/dev/sparrow/pkg/sparrow/targets/manager.go:177 msg="Registering as global target"
[...]

PR #122, opened 2024-03-07
refactor: genericize target manager

lvlcn-t:
## Motivation

This is the second PR to address #66.

Initially the target manager was built to implement the feature of #30 asap. The manager logic was implemented in a way around the gitlab API interactions. To be able to support other remote state backends for global registration we need to genericize the target manager.

## Changes

Pretty much everything is moving and renaming in this PR. 

The only thing I've really changed is that the config now has a type which specifies the type of the remote interactor used kind of like we already do for the loader.

The new target manager config looks like this:
```yaml
targetManager:
  type: gitlab # <- This is the new type config option.
  checkInterval: 1m
  registrationInterval: 1m
  updateInterval: 2m
  unhealthyThreshold: 3m
  gitlab:
    baseUrl: https://gitlab.com
    projectId: 123456
 ```

I'll change the docs and chart for this later with the last PR regarding #66 that will close the issue.

* refactor: generalize Gitlab interface to remote interactor interface
* refactor: mv gitlab config to interactor config
* chore: adjust imports
* feat: add interactor type
* feat: adjust target manager validation for interactor type
* chore: adjust naming
* chore: bump helm-docs v1.13.1

For additional information look at the commits.

## Tests done

- [x] Unit tests succeeded
- [ ] E2E tests succeeded - we have no e2e tests for the target manager

#### Logs
```
$ go run main.go run --config .tmp/config/start-config.yaml 
Using config file: .tmp/config/start-config.yaml
time=09:35:17 level=INFO source=/home/installadm/dev/github/sparrow/cmd/run.go:81 msg="Running sparrow"
time=09:35:17 level=INFO source=/home/installadm/dev/github/sparrow/pkg/sparrow/targets/manager.go:82 msg="Starting target manager reconciliation"
time=09:35:17 level=INFO source=/home/installadm/dev/github/sparrow/pkg/api/api.go:76 msg="Serving Api" addr=:8080
time=09:35:17 level=INFO source=/home/installadm/dev/github/sparrow/pkg/config/file.go:69 msg="File Loader disabled"
time=09:36:17 level=DEBUG source=/home/installadm/dev/github/sparrow/pkg/sparrow/targets/remote/gitlab/gitlab.go:142 msg="Fetching file list from gitlab"
time=09:36:17 level=DEBUG source=/home/installadm/dev/github/sparrow/pkg/sparrow/targets/remote/gitlab/gitlab.go:192 msg="Successfully fetched file list" files=3
[...]

PR #126, opened 2024-03-16
feat: handle os signals

lvlcn-t:
## Motivation

To address this conversation https://github.com/caas-team/sparrow/pull/125#discussion_r1525406062 

## Changes

We're now handling the os shutdown signals so the sparrow still shuts down, even in case of a controlled shutdown. This also addresses the bug that sparrow instances stay registered on the remote repository even though the sparrow instance doesn't run anymore.

Additionally to that I've adjusted the README to the changes made in #122 and removed the error handling of the (*Sparrow).Run method because it's always non nil.

- **feat: add os signal handling for graceful shutdown**
- **chore: housekeeping**

## Tests done

- [x] Unit tests succeeded
- [ ] E2E tests succeeded - we have none for graceful shutdown

## TODO


puffitos (approved):
LGTM

pkg/checks/traceroute/config.go
```diff
@@ -37,13 +37,12 @@ func (c *Config) Validate() error {
...
 		ip := net.ParseIP(t.Addr)
-
 		if ip != nil {
 			continue
 		}
 
 		_, err := url.Parse(t.Addr)
-		if err != nil && ip == nil {
+		if err != nil {
 			return checks.ErrInvalidConfig{CheckName: CheckName, Field: fmt.Sprintf("traceroute.targets[%d].addr", i), Reason: "invalid url or ip"}
```
puffitos, line 46: shouldn't this be err != nil || ip == nil, in case the IP is really nil after the `net.ParseIP`?

lvlcn-t: I've removed this because my linter complained and he's right. We're checking before if it's not nil:
https://github.com/caas-team/sparrow/blob/bed33794e600adc70b3e6a89f29557a7ddbdfe0a/pkg/checks/traceroute/config.go#L39-L42

So at this point `ip` is always nil.

PR #127, opened 2024-04-03
Push to MTR directly

puffitos:
## Motivation

We don't need to depend to MTR doing its mirroring for the image. We can directly push the image here once build and be done with it.

Mirroring is a deferred process, which hasn't been working lately (at least of the sparrow repository). Switching to a push based approach, given the small volume of traffic on this repository, is the simplest solution to move & test quickly.

## Changes

Changed the CI actions to push the OCI artrifacts to MTR as well.

## Tests done

None needed.

## TODO

- [x] Add secrets to repository

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,4 +101,4 @@ jobs:
       - name: Push helm package
         run: |
           helm push $(ls ./chart/*.tgz| head -1) oci://ghcr.io/${{ github.repository_owner }}/charts
-          helm push $(ls ./chart/*.tgz| head -1) oci://mtr.devops.telekom.de/sparrow/charts
+          helm push $(ls ./chart/*.tgz| head -1) oci://mtr.devops.telekom.de/sparrow/chart
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
         with:
           sarif_file: "trivy-results.sarif"
 
-      - name: GHRC Registry login
+      - name: GHCR login
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -24,11 +24,19 @@ jobs:
...
+        name: Login to GHCR
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - uses: docker/login-action@v3
+        name: Login to MTR
+        with:
+          registry: ghcr.io
```
y-eight, line 36: registry is not ghcr.io

y-eight (commented):
Problem here is that the sparrow repo is not in the caas orga in the MTR. The github MTR push user is set for the caas orga.
We need to set the secrets in the sparrow repo separately or push the image to the caas MTR orga instead.

.github/workflows/ci.yml
```diff
@@ -46,15 +46,26 @@ jobs:
         with:
           sarif_file: "trivy-results.sarif"
 
-      - name: Registry login
+      - name: GHRC Registry login
```
lvlcn-t, line 49: Typo

lvlcn-t (approved):
I don't really like the idea of pushing to our internal registry, but it makes sense with the recent problems of the MTR mirroring.

puffitos:
@y-eight it was a combination of multiple things... Typos, missing secrets, the go-releaser job doesn't allow one to add extra tags (the whole image, not just the tag in the end) so no mtr.devops.telekom.de image was available and so on. Now eveything works, with a small caveat: the charts are now being pushed to /charts/sparrow instead of just /chart (it's not possible AFAIK to define the repository, as helm always adds the name of the release to the repository you're uploading to)

@lvlcn-t I think that's not a problem. As long as MTR is public, I think it's something we may do. If it isn't anymore, we can remove the pushing and do some other kind of trick to mirror the images.

y-eight (approved):

PR #129, opened 2024-04-04
Bump chart's appVersion and fix chart repo on release job

puffitos:
## Motivation

The appVersion was false and the helm release job was pushing to a wrong repo.

## Changes

See commits.

## Tests done

None, seems pretty straightforward.

## TODO

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,4 +65,4 @@ jobs:
       - name: Push helm package
         run: |
           helm push $(ls ./chart/*.tgz| head -1) oci://ghcr.io/${{ github.repository_owner }}/charts
-          helm push $(ls ./chart/*.tgz| head -1) oci://mtr.devops.telekom.de/sparrow/chart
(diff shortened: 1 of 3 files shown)


lvlcn-t:
Can you also adjust the latest version in the README?

https://github.com/caas-team/sparrow/blob/main/README.md?plain=1#L84

puffitos:
> Can you also adjust the latest version in the README?
> 
> https://github.com/caas-team/sparrow/blob/main/README.md?plain=1#L84

Sorry, I don't think this is something we should do. Manual updates of the README, just to show how to install a specific version, don't really add any value to sparrow. It's nice updating everything and polishing the documentation every now and then, but this is where we should draw a line. Showing once how it's done is more than enough.

We already have to update the chart version manually each time we change something, and we need to remember to change the appVersion as well when we bump the sparrow's version. That's enough busywork for me.

y-eight (approved):

lvlcn-t (approved):
Yeah you're probably right, maybe we should automate this whole chore.

PR #130, opened 2024-04-06
chore/remove-controller-errors

### LINKED ISSUE #123: Code smell: no error returned in any controller function

### Current Behavior

During https://github.com/caas-team/sparrow/issues/37 I noticed that the ChecksController doesn't ever return an error in any of its functions, as they are all throttled down the error channel.

When the err is always nil, then it should not be part of the function's signature :)

### Expected Behavior

No error as a return parameter.

### Steps To Reproduce

_No response_

_No response_

Everyone

### Anything else?

_No response_

puffitos:
## Motivation

Closes #123

## Changes

Removed the errors from the controller functions, as they aren't/ produced at all.

## Tests done

- [x] Unit tests succeeded

## TODO

### DIFF
--- a/pkg/sparrow/controller.go
+++ b/pkg/sparrow/controller.go
@@ -78,7 +78,7 @@ func (cc *ChecksController) Run(ctx context.Context) error {
 }
 
 // Shutdown shuts down the ChecksController.
-func (cc *ChecksController) Shutdown(ctx context.Context) (err error) {
+func (cc *ChecksController) Shutdown(ctx context.Context) {
 	log := logger.FromContext(ctx)
 	log.Info("Shutting down checks controller")
 
@@ -88,7 +88,6 @@ func (cc *ChecksController) Shutdown(ctx context.Context) (err error) {
 	cc.done <- struct{}{}
 	close(cc.done)
 	close(cc.cResult)
-	return err
 }
 
 // Reconcile reconciles the checks.
@@ -124,7 +123,7 @@ func (cc *ChecksController) Reconcile(ctx context.Context, cfg runtime.Config) {
 
 	// Register new checks
 	for _, c := range newChecks {
-		err = cc.RegisterCheck(ctx, c)
+		cc.RegisterCheck(ctx, c)
 		if err != nil {
 			log.ErrorContext(ctx, "Failed to register check", "check", c.Name(), "error", err)
 		}
@@ -132,7 +131,7 @@ func (cc *ChecksController) Reconcile(ctx context.Context, cfg runtime.Config) {
 }
 
 // RegisterCheck registers a new check.
-func (cc *ChecksController) RegisterCheck(ctx context.Context, check checks.Check) error {
+func (cc *ChecksController) RegisterCheck(ctx context.Context, check checks.Check) {
 	log := logger.FromContext(ctx).With("check", check.Name())
 
 	// Add prometheus collectors of check to registry
@@ -154,7 +153,6 @@ func (cc *ChecksController) RegisterCheck(ctx context.Context, check checks.Chec
 	}()
 
 	cc.checks.Add(check)
-	return nil
 }
 
 // UnregisterCheck unregisters a check.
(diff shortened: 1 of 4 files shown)


pkg/sparrow/controller.go
```diff
@@ -124,15 +123,15 @@ func (cc *ChecksController) Reconcile(ctx context.Context, cfg runtime.Config) {
 
 	// Register new checks
 	for _, c := range newChecks {
-		err = cc.RegisterCheck(ctx, c)
+		cc.RegisterCheck(ctx, c)
 		if err != nil {
 			log.ErrorContext(ctx, "Failed to register check", "check", c.Name(), "error", err)
 		}
```
lvlcn-t, line 129: You can remove the error check here

pkg/sparrow/controller_test.go
```diff
@@ -236,69 +231,57 @@ func TestChecksController_Reconcile(t *testing.T) {
 
 func TestChecksController_RegisterCheck(t *testing.T) {
 	tests := []struct {
-		name    string
-		setup   func(t *testing.T) *ChecksController
-		check   checks.Check
-		wantErr bool
+		name  string
+		setup func(t *testing.T) *ChecksController
```
lvlcn-t, line 235: you can remove the `t *testing.T` argument here.

pkg/sparrow/controller_test.go
```diff
@@ -236,69 +231,57 @@ func TestChecksController_Reconcile(t *testing.T) {
...
 }
 
 func TestChecksController_UnregisterCheck(t *testing.T) {
 	tests := []struct {
-		name    string
-		setup   func(t *testing.T) *ChecksController
-		check   checks.Check
-		wantErr bool
+		name  string
+		setup func(t *testing.T) *ChecksController
```
lvlcn-t, line 269: > you can remove the `t *testing.T` argument here.

Here aswell

pkg/sparrow/run.go
```diff
@@ -172,7 +172,7 @@ func (s *Sparrow) shutdown(ctx context.Context) {
 		}
 		sErrs.errAPI = s.api.Shutdown(ctx)
 		s.loader.Shutdown(ctx)
-		sErrs.errCheckCont = s.controller.Shutdown(ctx)
```
lvlcn-t, line 175: You can also remove the field from the error struct.

lvlcn-t (changes requested):
Some minor things

pkg/sparrow/controller_test.go
puffitos: I've removed the arguments from one of the tests and I've completely removed the function in the other test, as it was returning the same thing in both cases of the test suite.

pkg/sparrow/controller_test.go
```diff
@@ -236,79 +231,60 @@ func TestChecksController_Reconcile(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			cc := tt.setup(t)
+			cc := tt.setup()
 
-			err := cc.RegisterCheck(context.Background(), tt.check)
-			if (err != nil) != tt.wantErr {
-				t.Errorf("RegisterCheck() error = %v, wantErr %v", err, tt.wantErr)
-			}
+			cc.RegisterCheck(context.Background(), tt.check)
```
lvlcn-t, line 261: We should check for something in this test, otherwise we only test if it runs succussfully

puffitos: @lvlcn-t The scope of this PR is to remove the error from the controller. Let's not change the tests for now (which are fragile at best). We can open an issue regarding the tests.

lvlcn-t: Yeah, let's open an issue.

lvlcn-t (approved):

pkg/sparrow/controller_test.go
puffitos: https://github.com/caas-team/sparrow/issues/131 done, will post PR soon to fix

PR #159, opened 2024-07-29
Feat/improve traceroute

niklastreml:
## Motivation

This PR aims to improve the traceroute check. 

### Feature: Retrieves IP addresses of routers when icmp is available
We have now removed the dependency on `github.com/aeden/traceroute` by building our own implementation. This gives us more flexibility in how the check runs, mainly allowing us to heavily parallelize it and instrument it. The core of this feature can be found in [`traceroute.go`](https://github.com/caas-team/sparrow/blob/906521cb0e5853120597ad4c75daf9da5ef19477/pkg/checks/traceroute/traceroute.go)
### Performance: Concurrency for everything
Previously, the check would run every target and every hop of that target serially, causing a single cycle, even with low amounts of targets and hop counts, to take time on the order of multiple minutes. 

This pr addresses this, by running all targets and their hops in parallel. For this to work properly, we need to ensure that we don't mix up ICMP of different execution threads, that are coming in at the same time. This is easily done by reading the payload of the incoming ICMP `Time Exceeded` packet. The payload of this packet will include the ip header + 64 bits of the payload of the packet that caused this `Time Exceeded` message. Those 64 bits just about contain the tcp packets source and destination ports. We use this functionality by assigning a random portnumber to every thread, relying on the OS to ensure that no two threads have the same port. When an ICMP packet comes we check the payload for the port and discard it, if it doesn't match the threads assigned port. For now every thread gets its own ICMP socket, to keep logic simple. If this causes issues in the future we should refactor it, so there is only one shared socket.

### Fix: Shutdown hangs forever
The pr addresses an issue in the checks shutdown logic, where an unbuffered channel caused the checks shutdown routine to block forever.

### Metrics: Prometheus metrics
It doesn't make sense to try and use prometheus to capture any of the more detailed traceroute metrics since those would be of pretty high cardinality and would essentially cause prometheus to explode. 

I've still added some metrics, that I thought made sense:
```console
$ curl -s localhost:8080/metrics | grep traceroute
# HELP sparrow_traceroute_check_duration_ms
# TYPE sparrow_traceroute_check_duration_ms gauge
sparrow_traceroute_check_duration_ms{target="google.com"} 415
sparrow_traceroute_check_duration_ms{target="telekom.de"} 232
# HELP sparrow_traceroute_minimum_hops
# TYPE sparrow_traceroute_minimum_hops gauge
sparrow_traceroute_minimum_hops{target="google.com"} 14
sparrow_traceroute_minimum_hops{target="telekom.de"} 4
```

### Metrics: API Metrics
There are more detailed metrics available through the json API, which roughly matches the output provided by traceroute when run in a terminal:
```console
❯ curl -s localhost:8080/v1/metrics/traceroute | head -n18
{
  "data": {
    "telekom.de": {
      "MinHops": 4,
      "Hops": {
        "1": [
          {
            "Latency": 14102685,
            "Addr": {
              "IP": "10.74.11.36",
              "Port": 80,
              "Zone": ""
            },
            "Name": "telekom.de.",
            "Ttl": 1,
            "Reached": false
          }
        ],
    ...
```

### Development: Added a local test setup for traceroute
To be able to actually test this thing atleast manually I've provided a [kathara](https://github.com/KatharaFramework/Kathara)-based test setup, which is documented [here](https://github.com/caas-team/sparrow/blob/906521cb0e5853120597ad4c75daf9da5ef19477/pkg/checks/traceroute/test-lab/how-to-test.md). This should make it pretty easy to test and debug this. We can maybe even use this tool for e2e tests in CI

### Testing: Added E2E tests for the traceroute check
This test sets up a kathara network and starts sparrow on one system, tracerouting to the others webserver. It then checks the prometheus metrics and the metrics api to ensure that the data collected is what is expected. Look at [this](https://github.com/caas-team/sparrow/actions/runs/10179635177/job/28155787157?pr=159#step:6:1) for some example output.
### Guide for reviewing
I have moved the logic of the check (setup, shutdown, implementing the interface and such) into `check.go` (prev. `traceroute.go`). `traceroute.go` now contains the implementation of the actual logic. In general I tried to keep my changes constrained to that file, by only building a new implementation of the [`tracerouteFactory`](https://github.com/caas-team/sparrow/blob/906521cb0e5853120597ad4c75daf9da5ef19477/pkg/checks/traceroute/check.go#L52) type, so as to make reviewing the code easier. It's probably best to review `traceroute.go` completely start to finish, ignoring the diff on github and diffing the `check.go` against `traceroute.go` of the main branch. 

### Setup opentelemetry
We want to expose the more detailed in a standardized format, that easily be consumed by tools like grafana. OpenTelemetry should be fine and simple to implement, as the go-sdk is nice to use and the protocol is also well supported. Once that's doen we can create otel traces for the traceroute hops, that can be visualized easily.

### Support ICMP and UDP
Currently we only support TCP as the probing protocol. After this is merged, I will do a small refactor and make the probing protocol swappable. UDP should be simple to implement (basically just change the string in the constructor of the dialer, the rest of the logic is prety similar to TCP). ICMP might be a little tricky, since we need to ensure, that we can differentiate packets sent by different threads, like tcp, but this time, we can't rely on the OS to ensure no duplicate ports being used.

## Tests done

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO


pkg/checks/traceroute/check.go
```diff
@@ -0,0 +1,195 @@
...
+			l.Debug("Ran traceroute", "result", trace, "duration", elapsed)
+			r := result{
+				Hops:    trace,
+				MinHops: tr.config.MaxHops,
+			}
+
+			for i, hops := range trace {
+				for _, hop := range hops {
+					if hop.Reached && hop.Ttl < r.MinHops {
+						r.MinHops = i
```
y-eight, line 140: ~~r.MinHps should be i-1~~
 
edit: ok since ttl is obviously starting at 1 ✅

y-eight: ~~Is there a possibility to have multiple hops per hop since hops is a slice as well?~~

edit: sure! retries are listed in on hop.

maybe the wording here is not correct. would it be better to use: `hop` instead of `hops` and `retry` instead of `hop`

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,263 @@ package traceroute
...
 
-// Run runs the check in a loop sending results to the provided channel
-func (tr *Traceroute) Run(ctx context.Context, cResult chan checks.ResultDTO) error {
-	ctx, cancel := logger.NewContextWithLogger(ctx)
-	defer cancel()
-	log := logger.FromContext(ctx)
+// TraceRoute performs a traceroute to the specified host using TCP and listens for ICMP Time Exceeded messages using ICMP.
+func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error) {
+	// this could also be a 2d array, but I feel like using an int map here makes the json easier to understand
+	// as it explicitly shows a mapping of ttl->hops
```
y-eight, line 93: fine... pls change the comment here sth like `mapping of a specific hop to its ttl (map key)`

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,263 @@ package traceroute
...
-	type internalResult struct {
-		addr string
-		res  result
+	addr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:%d", cfg.Dest, cfg.Port))
+	if err != nil {
+		log.Error("failed to resolve target name", "err", err.Error())
+		return nil, err
 	}
 
+	results := make(chan Hop, cfg.MaxHops)
```
y-eight, line 105: Mhm, I don't like that channel construct here but I don't have a better idea so far.

But nevertheless, the channel should have a size of `cfg.MaxHops * cfg.Rc` to provide an appropriate size in case of retrying at least one hop and max out the maxHops.

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,263 @@ package traceroute
...
-// This should return an error if the config is invalid
-func (tr *Traceroute) SetConfig(cfg checks.Runtime) error {
-	if cfg, ok := cfg.(*Config); ok {
-		tr.Mu.Lock()
-		defer tr.Mu.Unlock()
-		tr.config = *cfg
-		return nil
-	}
+		// Check if the destination port matches our dialer's source port
+		if canIcmp && gotPort == clientPort {
```
y-eight, line 212: You could simplify this by using:

```golang
if !canIcmp {
  break
}
```

y-eight: I think this func should be slitted, its hard to read and to reduce complexity. Maybe a func `watchForIcmp` (or sth else) can be used that is called when initial tcpHop has failed

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,263 @@ package traceroute
...
+	case *net.TCPAddr:
+		return addr.IP
+	case *net.IPAddr:
+		return addr.IP
 	}
+	return nil
+}
 
-	log.Debug("Waiting for traceroute checks to finish")
+func traceroute(results chan Hop, addr net.Addr, ttl int, timeout time.Duration) (bool, error) {
```
y-eight, line 154: add comment for docu please

y-eight (changes requested):
First review done

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,327 @@ package traceroute
...
+	Ttl     int           `json:"ttl" yaml:"ttl" mapstructure:"ttl"`
+	Reached bool          `json:"reached" yaml:"reached" mapstructure:"reached"`
 }
 
-// GetMetricCollectors allows the check to provide prometheus metric collectors
-func (tr *Traceroute) GetMetricCollectors() []prometheus.Collector {
-	return []prometheus.Collector{}
+type HopAddress struct {
+	IP   string `json:"ip" yaml:"ip" mapstructure:"ip"`
+	Port int    `json:"port" yaml:"port" mapstructure:"port"`
```
lvlcn-t, line 300: I think it's appropriate to use a `uint16` type here

pkg/checks/traceroute/check.go
```diff
@@ -0,0 +1,195 @@
...
+// Shutdown is called once when the check is unregistered or sparrow shuts down
+func (tr *Traceroute) Shutdown() {
+	tr.DoneChan <- struct{}{}
+	close(tr.DoneChan)
+}
+
+// SetConfig is called once when the check is registered
+// This is also called while the check is running, if the remote config is updated
+// This should return an error if the config is invalid
+func (tr *Traceroute) SetConfig(cfg checks.Runtime) error {
```
lvlcn-t, line 168: Maybe we should log a warning if the binary has no capabilities set to signalize the user as soon as he wants to use the check that he may have forgot to set the capabilities. To check for capabilities, you can refer to my [implementation](https://github.com/caas-team/sparrow/blob/feat/tcp-traceroute/internal/helper/capabilities.go#L16) of the traceroute check:
```go
// Capability represents a Linux capability
type Capability uint32

const (
	// CAP_NET_RAW is the capability to use raw sockets
	CAP_NET_RAW Capability = (1 << uint32(13))
)

// HasCapabilities checks if the current process has the specified capabilities.
func HasCapabilities(ca Capability) bool {
	if unix.Geteuid() == 0 {
		return true
	}

	hdr := &unix.CapUserHeader{
		Pid:     int32(unix.Getpid()),
		Version: unix.LINUX_CAPABILITY_VERSION_3,
	}
	data := &unix.CapUserData{}

	err := unix.Capget(hdr, data)
	if err != nil {
		return false
	}

	return data.Effective&uint32(ca) != 0
}
```

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,327 @@ package traceroute
...
-func newTraceroute(dest string, port, timeout, retries, maxHops int) (traceroute.TracerouteResult, error) {
-	opts := &traceroute.TracerouteOptions{}
-	opts.SetTimeoutMs(timeout)
-	opts.SetRetries(retries)
-	opts.SetMaxHops(maxHops)
-	opts.SetPort(port)
-	return traceroute.Traceroute(dest, opts)
-}
+	// The first 20 bytes of Data are the IP header, so the TCP segment starts at byte 20
+	tcpSegment := msg.Body.(*icmp.TimeExceeded).Data[20:]
```
lvlcn-t, line 85: Consider adding a constant for this, so it's clearer when reading just the code.

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,327 @@ package traceroute
...
-func (tr *Traceroute) Name() string {
-	return CheckName
+func (a HopAddress) String() string {
+	if a.Port != 0 {
+		return fmt.Sprintf("%s:%d", a.IP, a.Port)
+	}
+	return a.IP
+}
+
+func printHops(ctx context.Context, mapHops map[int][]Hop) {
```
lvlcn-t, line 310: Consider renaming this to `logHops`

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,327 @@ package traceroute
...
+	}
 
-	for r := range cResult {
-		res[r.addr] = r.res
+	h := handleIcmpResponse(ctx, icmpListener, clientPort, ttl, timeout)
+	h.Latency = latency
+	return &h, nil
+}
+
+func setupIcmpListener() (bool, *icmp.PacketConn, error) {
```
lvlcn-t, line 194: Please rename this to `newICMPListener` to match the naming patterns of the sparrow.

.github/workflows/e2e_checks.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Build binary for e2e
+        uses: goreleaser/goreleaser-action@v5
+        with:
+          version: latest
+          args: build --single-target --clean  --snapshot --config .goreleaser-ci.yaml
+
+      - name: Run e2e tests
+        run: |
+          ./scripts/run_e2e_tests.sh
```
lvlcn-t, line 54: Can you set the environment variable `LOG_LEVEL` to `debug` here so we can directly see all logs?

lvlcn-t (commented):
First review, overall good implementation only some minor things to think about/improve 👍

pkg/checks/traceroute/check.go
niklastreml: I decided against doing a check like this, since I'm not 100% sure if theyre are edge cases where this would return false negative. I don't know if those are the only two ways of checking if the process is allowed CAP_NET_RAW. To me it feels safer to just ask the os directly: "Can I open this socket" instead. That way we don't have to try and mirror the kernels authz logic

pkg/checks/traceroute/traceroute.go
niklastreml: Why? this really doesn't matter

pkg/checks/traceroute/traceroute.go
niklastreml: I didn't use a uint16, because the [standard](https://cs.opensource.google/go/go/+/go1.22.5:src/net/tcpsock.go;l=21) [library](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/udpsock.go;l=24) doesn't either:

```go
type UDPAddr struct {
	IP   IP
	Port int
	Zone string // IPv6 scoped addressing zone
}

// TCPAddr represents the address of a TCP end point.
type TCPAddr struct {
	IP   IP
	Port int
	Zone string // IPv6 scoped addressing zone
}
```

pkg/checks/traceroute/traceroute.go
```diff
@@ -2,197 +2,327 @@ package traceroute
...
-func (tr *Traceroute) GetConfig() checks.Runtime {
-	tr.Mu.Lock()
-	defer tr.Mu.Unlock()
-	return &tr.config
-}
-
-func (tr *Traceroute) check(ctx context.Context) map[string]result {
-	res := make(map[string]result)
-	log := logger.FromContext(ctx)
+	timeoutDuration := time.Duration(cfg.Timeout) * time.Second
```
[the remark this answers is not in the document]
niklastreml: It is

lvlcn-t (approved):
LGTM

PR #180, opened 2024-09-05
Fix/remove labelled metrics

### LINKED ISSUE #77: Bug: Removed targets remain as prometheus metrics

### Current Behavior

Sparrow starts with two targets:
```
https://google.com
https://example.com
```

this creates two prometheus for those targets:
```
sparrow_latency_duration_seconds{status="200",target="https://google.com"} 0.014013145
sparrow_latency_duration_seconds{status="200",target="https://example.com"} 0.003617327
...
```
If the config gets updated and one target gets removed, the last prometheus metric for the deleted target remains as the latest prometheus metrics. 
New targets:
```
https://google.com
```
Prometheus metrics:
```
sparrow_latency_duration_seconds{status="200",target="https://google.com"} 0.014013145 <- This remains and is not cleaned up
sparrow_latency_duration_seconds{status="200",target="https://example.com"} 0.007915321
```

### Expected Behavior

Lets discuss how (if) we should handle this case.

Maybe it's possible to remove metrics from the prometheus registry

### Steps To Reproduce

1. Deploy sparrow with two targets for any check using http loader
2. Let it run for a check cycle
3. Remove one target from the config
4. HTTP GET /metrics

Screenshot shows what happened after google.com was removed from the config, compared to a target that is currently active

image

Since the sparrow metrics api still exposes the value for google.com, prometheus stores it on every scrape

@puffitos @y-eight @lvlcn-t Let's discuss how we handle this next week

### Anything else?

_No response_

puffitos:
## Motivation

Closes #77 

## Changes

- Remove non-used metrics when setting the configuration (change the name of the interface method from `SetConfig` to `UpdateConfig`)
  - what happens when the config gets set and how the metrics are supposed to be cleaned up should be a responsibility of the check itself. Placing that responsibility in the CheckController feels wrong and forces us to write a function with a switch / if-else, depending on what kind of `Config` is being passed, since `Config` is just an interface.
  - This adds the drawback of having to implement 2 more funtions per Check (one interface function and an almost necessary remove function for the metrics struct, which can be implemented as the case needs it to.
  - Adding a function to the interface isn't a must, we could still change this to avoid inflating the interface. Still, since our checks all must have a target (no other kind of check is supported / is currently in scope), my first gut feeling is that it's OK :)
- Removed the deprecated metric, as it's been marked for removal since https://github.com/caas-team/sparrow/releases/tag/v0.4.0 (6/6)

As a result:

- The prometheus metrics are also displayed correctly. The old ones are still in the database, but after removing the target no value is produced anymore, as is wanted/ expected/ reasonable/ needed to solve the issue.

## Tests done

- [x] Unit tests succeeded
- [x] Manual E2E tests succeeded

### Manual tests done

- ✅ (for each checks outside traceroute, one by one) Added new target → target was added to metric → target is present in metric → target removed from config → target is removed from metrics
- ✅ (all checks at once)  Added new target to all metrics → target was added to metrics → target is present in metrics → target removed from all metrics in config → target is removed from all metrics

## TODO


pkg/checks/base.go
```diff
@@ -56,6 +56,9 @@ type Check interface {
 	Schema() (*openapi3.SchemaRef, error)
 	// GetMetricCollectors allows the check to provide prometheus metric collectors
 	GetMetricCollectors() []prometheus.Collector
+	// RemoveLabelledMetrics allows the check to remove prometheus metric collectors
```
y-eight, line 59: This is not for rm the metric collectors. It's for removing metrics by target from available collectors

y-eight (changes requested):
I will check +with Marius from SRE) whether they use the rm deprecated label anywhere.

pkg/checks/base.go
puffitos: Yes, the metrics struct in prometheus implements the collector interface, that's why I used the name collector... The GaugeVec, for example, is just a `Collector`, as everything else. The naming of the `GetMetricCollectors` is also not the best. :shrug:

I'll just change the naming in the comment. I'm still unsure on whether the interface is the proper place for metrics handling...

pkg/checks/base.go
```diff
@@ -56,6 +56,9 @@ type Check interface {
 	Schema() (*openapi3.SchemaRef, error)
 	// GetMetricCollectors allows the check to provide prometheus metric collectors
 	GetMetricCollectors() []prometheus.Collector
+	// RemoveLabelledMetrics allows the check to remove the prometheus metrics
+	// of the check whose `target` label matches the passed value
+	RemoveLabelledMetrics(target string) error
 }
```
lvlcn-t, line 62: I think we should split this interface up into multiple interfaces that can be implemented by a check in the near future. 

For example if a check cannot expose any prometheus metrics, it should not need to implement these methods. I think the best way forward would be to move everything out of the `Check` interface and then composing the check by using the new smaller interfaces:
```go
// Runner interface for running and shutting down the check
type Runner interface {
	Run(ctx context.Context, cResult chan ResultDTO) error
	Shutdown()
}

// Configurable interface for setting and getting configuration
type Configurable interface {
	UpdateConfig(config Runtime) error
	GetConfig() Runtime
}

// Identifiable interface for getting the name of the check
type Identifiable interface {
	Name() string
}

// SchemaProvider interface for providing schema
type SchemaProvider interface {
	Schema() (*openapi3.SchemaRef, error)
}

// MetricProvider interface for providing metric collectors
type MetricProvider interface {
	GetMetricCollectors() []prometheus.Collector
	RemoveLabelledMetrics(target string) error
}

// Check implementations are expected to perform specific monitoring tasks and report results.
type Check interface {
	Runner
	Configurable
	Identifiable
	SchemaProvider
	MetricProvider
}
```

This way we can have a more flexible interface that can be implemented by checks that do not need to implement all the methods.

pkg/checks/errors.go
```diff
@@ -42,3 +42,12 @@ type ErrInvalidConfig struct {
...
 	return fmt.Sprintf("invalid configuration field %q in check %q: %s", e.Field, e.CheckName, e.Reason)
 }
+
+// ErrMetricNotFound is returned when a metric is not found
+type ErrMetricNotFound struct {
+	Label string
+}
+
+func (e ErrMetricNotFound) Error() string {
+	return fmt.Sprintf("metric not found: %s", e.Label)
```
lvlcn-t, line 52: Consider using the `%q` verb instead

pkg/checks/latency/latency.go
```diff
@@ -243,8 +205,6 @@ func (l *Latency) check(ctx context.Context) map[string]result {
 			mu.Lock()
 			defer mu.Unlock()
 
-			l.metrics.duration.DeletePartialMatch(prometheus.Labels{"target": target})
-			l.metrics.duration.WithLabelValues(target, strconv.Itoa(results[target].Code)).Set(results[target].Total)
 			l.metrics.totalDuration.WithLabelValues(target).Set(results[target].Total)
 			l.metrics.count.WithLabelValues(target).Inc()
```
lvlcn-t, line 209: Can you take a look here? I think we increment this twice for each target which results in an incorrect count.

pkg/checks/base.go
puffitos: That looks like a good idea, at first glance. I'll take a swing at it and see where it goes. But not as part of this PR, as this would totally be outta scope. 😉

pkg/checks/latency/latency.go
puffitos: Indeed:

```shell
# HELP sparrow_latency_count Count of latency checks done
# TYPE sparrow_latency_count counter
sparrow_latency_count{target="https://gitlab.devops.telekom.de"} 7
sparrow_latency_duration_count{target="https://gitlab.devops.telekom.de"} 4
```

And now with the change, the two metrics (which should always have the same value) go hand-in-hand:

```shell
# HELP sparrow_latency_count Count of latency checks done
# TYPE sparrow_latency_count counter
sparrow_latency_count{target="https://gitlab.devops.telekom.de"} 9
sparrow_latency_duration_count{target="https://gitlab.devops.telekom.de"} 9
```

All checks now have the same count, as expected (when the interval of each check is the same):


```shell
# HELP sparrow_dns_check_count Total number of DNS checks performed on the target and if they were successful.
# TYPE sparrow_dns_check_count counter
sparrow_dns_check_count{target="gitlab.devops.telekom.de"} 13
sparrow_dns_duration_count{target="gitlab.devops.telekom.de"} 13
# HELP sparrow_latency_count Count of latency checks done
# TYPE sparrow_latency_count counter
sparrow_latency_count{target="https://gitlab.devops.telekom.de"} 13
sparrow_latency_duration_count{target="https://gitlab.devops.telekom.de"} 13
```

pkg/checks/latency/latency.go
```diff
@@ -71,7 +71,6 @@ type result struct {
 
 // metrics defines the metric collectors of the latency check
 type metrics struct {
-	duration      *prometheus.GaugeVec
 	totalDuration *prometheus.GaugeVec
 	count         *prometheus.CounterVec
 	histogram     *prometheus.HistogramVec
```
lvlcn-t, line 76: Please move the metrics struct to the `pkg/checks/latency/metrics.go` file so everything is always in the same location.

pkg/checks/health/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GetMetricCollectors returns all metric collectors of check
+func (h *Health) GetMetricCollectors() []prometheus.Collector {
+	return []prometheus.Collector{
+		h.metrics,
+	}
+}
+
+// Remove removes a metric with a specific label
+func (m metrics) Remove(label string) error {
```
lvlcn-t, line 36: Is there a reason why you've chosen not to use a pointer receiver here or was this just unintended?

pkg/checks/traceroute/check.go
```diff
@@ -49,14 +50,16 @@ type tracerouteConfig struct {
...
-}
+type (
+	tracerouteFactory func(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error)
+	result            struct {
+		// The minimum number of hops required to reach the target
+		MinHops int `json:"min_hops" yaml:"min_hops" mapstructure:"min_hops"`
+		// The path taken to the destination
+		Hops map[int][]Hop `json:"hops" yaml:"hops" mapstructure:"hops"`
+	}
+)
```
lvlcn-t, line 62: Imo this syntax is rather ugly for declaring multiple types but I care too little about it to be annoyed from it, so keep it this way if you like it better.

puffitos: golangci lint takes care of formatting. I'm just a mindless drone. ;)

pkg/checks/health/metrics.go
puffitos: Unintended. Should be a pointer, just for consistency's sake.

pkg/checks/latency/latency.go
puffitos: Done

lvlcn-t (approved):
LGTM

y-eight (approved):

PR #186, opened 2024-09-10
feat: instrument traceroute check

lvlcn-t:
## Motivation

Partly addresses #111 but could also be extended to the other checks and parts of the sparrow to provide technical information about the sparrow.

## Changes

I've instrumented the traceroute check to provide more information about how the traceroute performed. The traceroute check now starts a span for each target to trace the route to. In this span each hop then starts a new span to provide information about the hop. The spans are then sent to the configured telemetry backend.

## Tests done

- [ ] ~Unit tests succeeded~ - Not applicable for this PR
- [x] E2E tests succeeded

### Manual e2e tests

Here is how the traces will then look like in Jaeger:

image

And this is how a trace then looks like:

image

## TODO


pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
 	var wg sync.WaitGroup
 
 	for ttl := 1; ttl <= cfg.MaxHops; ttl++ {
+		c, hopSpan := tr.Start(ctx, fmt.Sprintf("Hop %d", ttl), trace.WithAttributes(
```
y-eight, line 179: The most important information about the hop is the address. Lets use the address (ip + port) here instead of `Hop X`.

pkg/checks/traceroute/check.go
```diff
@@ -32,6 +41,7 @@ func NewCheck() checks.Check {
 		config:     Config{},
 		traceroute: TraceRoute,
 		metrics:    newMetrics(),
+		tracer:     otel.Tracer("tracer.traceroute"),
```
puffitos, line 44: let's use the `CheckName` here? I don't think there will be any conflicts with other traces, as each check will have the one.

pkg/checks/traceroute/traceroute.go
```diff
@@ -38,9 +41,14 @@ func randomPort() int {
 	return rand.N(portRange) + basePort // #nosec G404 // math.rand is fine here, we're not doing encryption
 }
 
-func tcpHop(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration) (net.Conn, int, error) {
+// tcpHop attempts to connect to the target host using TCP with the specified TTL and timeout.
+// It returns a [net.Conn], the port used for the connection, and an error if the connection failed.
+func tcpHop(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration) (conn net.Conn, port int, err error) {
```
puffitos, line 46: Is there really a need for the named returns? If there really is, then we should consider refactoring this function. Seems OK without the named returns.

pkg/checks/traceroute/traceroute.go
```diff
@@ -58,15 +66,46 @@ func tcpHop(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration)
...
+			// we are expecting the connection to fail because of TTL expiry
+			span.SetStatus(codes.Unset, "No route to host")
+			span.AddEvent("No route to host", trace.WithAttributes(
+				attribute.String("error", err.Error()),
+			))
+			logger.FromContext(ctx).DebugContext(ctx, "No route to host", "error", err.Error())
 			return conn, port, err
 		}
+
+		span.AddEvent("TCP connection failed", trace.WithAttributes(
```
puffitos, line 100: Can we move this to the default case, as we're already doing the error handling there?

pkg/checks/traceroute/traceroute.go
```diff
@@ -58,15 +66,46 @@ func tcpHop(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration)
...
+			continue
+		case errors.Is(err, syscall.EHOSTUNREACH):
+			// No route to host is no error because of how tcp traceroute works
+			// we are expecting the connection to fail because of TTL expiry
+			span.SetStatus(codes.Unset, "No route to host")
+			span.AddEvent("No route to host", trace.WithAttributes(
+				attribute.String("error", err.Error()),
+			))
+			logger.FromContext(ctx).DebugContext(ctx, "No route to host", "error", err.Error())
 			return conn, port, err
```
puffitos, line 97: Is this an error or not? I'm confused now :D If it's not an error, then please return nil as the third argument. If it's an error, then we mark should mark the span as a faulty one.

pkg/checks/traceroute/traceroute.go
```diff
@@ -109,13 +148,24 @@ func readIcmpMessage(ctx context.Context, icmpListener *icmp.PacketConn, timeout
...
+	defer sp.End()
+
+	// Maps ttl -> attempted hops for that ttl
 	hops := make(map[int][]Hop)
 	log := logger.FromContext(ctx).With("target", cfg.Dest)
 
 	addr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:%d", cfg.Dest, cfg.Port))
 	if err != nil {
-		log.Error("failed to resolve target name", "err", err.Error())
+		sp.RecordError(err)
```
puffitos, line 167: `SetStatus` is missing here, so we know that this span (and not only the child spans) has an error.

pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
...
 		go func(ttl int) {
 			defer wg.Done()
+			defer hopSpan.End()
+
 			l := log.With("ttl", ttl)
-			logctx := logger.IntoContext(ctx, l)
+			logctx := logger.IntoContext(c, l)
+
+			retry := 0
 			err := helper.Retry(func(ctx context.Context) error {
```
puffitos, line 191: `err` redeclaration here - I think we can just assign.

pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
...
+				))
+
 				hop, err := traceroute(ctx, addr, ttl, cfg.Timeout)
 				if hop != nil {
 					results <- *hop
 				}
 				if err != nil {
-					l.Error("traceroute failed", "err", err.Error())
+					l.ErrorContext(ctx, "Failed to trace route", "err", err.Error())
+					hopSpan.RecordError(err)
```
puffitos, line 206: `SetStatus` should be set here as well.

pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
...
 			err := helper.Retry(func(ctx context.Context) error {
+				defer func() {
+					retry++
+				}()
+				hopSpan.AddEvent("Attempting to hop", trace.WithAttributes(
+					attribute.Int("ttl", ttl),
+					attribute.Int("retry", retry),
+				))
+
 				hop, err := traceroute(ctx, addr, ttl, cfg.Timeout)
```
puffitos, line 200: `err` redeclared here, please rename to `ierr` or something similar. So many ifs and errors here, I don't even want to spend time thinking whether we're safe here and won't just return a wrong error...   

Please enlighten me if I'm mistaken.

pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
...
 				if hop != nil {
 					results <- *hop
 				}
 				if err != nil {
-					l.Error("traceroute failed", "err", err.Error())
+					l.ErrorContext(ctx, "Failed to trace route", "err", err.Error())
+					hopSpan.RecordError(err)
 					return err
 				}
 				if !hop.Reached {
```
puffitos, line 209: Is there a reason for no tracing error here?

pkg/checks/traceroute/traceroute.go
```diff
@@ -126,44 +176,69 @@ func TraceRoute(ctx context.Context, cfg tracerouteConfig) (map[int][]Hop, error
...
-					l.Debug("failed to reach target, retrying")
+					l.DebugContext(ctx, "failed to reach target, retrying")
 					return errors.New("failed to reach target")
 				}
 				return nil
 			}, cfg.Rc)(logctx)
 			if err != nil {
```
puffitos, line 215: Is there no need to record the error here as well? I would argue that if the target couldn't be reached, the span should display this information as well with `RecordError`

puffitos (commented):
Some questions regarding missing error spans mostly.

pkg/checks/traceroute/traceroute.go
lvlcn-t: I've just added them for documentation but if you think it makes it less readable, I can take them out again.

pkg/checks/traceroute/traceroute.go
lvlcn-t: Idk tbh, I've just use a new variable to be safe.

pkg/checks/traceroute/traceroute.go
lvlcn-t: Weird that the linter doesn't complain about it but thanks 🙂

pkg/checks/traceroute/traceroute.go
puffitos: We've already discussed this in previous occasions; named returns are a gotcha in go we don't want to mess with. Why roll the dice when there's no need to? The function is already documented. I'll let @niklastreml decide on that, as he's written this code.

Please keep in mind, that we're not using named returns elsewhere in the codebase. Consistency is also an important aspect of a solid codebase ;)

pkg/checks/traceroute/traceroute.go
puffitos: No need for a custom error, as special errors are typical in file and network operations.

Since the hop hasn't succeeded and there is indeed an error, shouldn't we mark the span as an error span as well? We need to mark each time an error happens, so it's easier to read the spans in the backend UI tool displaying them (Jaeger for example).

pkg/checks/traceroute/traceroute.go
puffitos: Tracing can be used to identify hard-to-solve errors and offer a debugging-like capability when it's impossible to run debug mode in production. IMO if an error happens, we should show it. It's important to know exactly where the error happens and not have to dig through all the tracing spans.

pkg/checks/traceroute/traceroute.go
```diff
@@ -109,105 +149,159 @@ func readIcmpMessage(ctx context.Context, icmpListener *icmp.PacketConn, timeout
...
-		return addr.IP
-	}
-	return nil
-}
-
-// traceroute performs a traceroute to the given address with the specified TTL and timeout.
+// hop performs a hop to the given address with the specified TTL and timeout.
 // It returns a Hop struct containing the latency, TTL, address, and other details of the hop.
-func traceroute(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration) (*Hop, error) {
+func hop(ctx context.Context, addr net.Addr, ttl int, timeout time.Duration) (*Hop, error) {
```
puffitos, line 248: Shouldn't we rename this to `doHop` or something similar to avoid the shadowing with the `hop` variable throughout the whole file? Or just use `res` or `h` for the hop in functions?

A conflict seems kind of impossible, but why even take the risk?

pkg/checks/traceroute/traceroute.go
puffitos: Fine by me :+1:

pkg/checks/traceroute/traceroute.go
niklastreml: Let's remove the named returns, the comments on the function should be documentation enough.

pkg/checks/traceroute/traceroute.go
lvlcn-t: I've changed this to be displayed as error instead of `codes.Unset`

niklastreml (approved):

y-eight (approved):

PR #212, opened 2024-10-16
fix: allow tls without a custom certificate

### LINKED ISSUE #211: Bug: TLS for telemetry is not possible, without specifying certs

### Current Behavior

It's not possible to use TLS for the otel exporter without passing a `telemetry.certPath`. This means impossible (or atleast annoying) to use the system's certificates.

### Expected Behavior

Sparrow should have an option like `telemetry.tls.enabled`. 
  
  If true sparrow should use the systems certificates, unless telemetry.tls.certPath is specified
 
  Otherwise use the `WithInsecure()` option

### Steps To Reproduce

1. Use this config:
```bash
cat config.yaml
```
```yaml
name: sparrow.localhost
telemetry:
  # exporter: stdout
  exporter: http
  url: some-https-jaeger.com
loader:
  type: file
  interval: 30s
  file:
    path: ./config.yaml
traceroute:
  interval: 1s
  timeout: 5s
  retries: 3
  maxHops: 32
  targets:
    - addr: telekom.de
      port: 443
```
2. `sparrow run --config config.yaml`

[Code](https://github.com/caas-team/sparrow/blob/main/pkg/sparrow/metrics/exporter.go#L94-L109)

Anyone

### Anything else?
This should be a simple fix, since we just have to add another boolean option to the config and then optionally allow passing a certpath

niklastreml:
## Motivation

Fixes #211 
## Changes

Added a new option for enabling TLS without custom certs. I changed the config file a bit for this, grouping the tls stuff into their own object. Check the readme for more info

## Tests done

- [x] Tested against jaeger instance exposed over https
## TODO

### DIFF
--- a/pkg/sparrow/metrics/exporter.go
+++ b/pkg/sparrow/metrics/exporter.go
@@ -122,12 +122,13 @@ func newGRPCExporter(ctx context.Context, config *Config) (sdktrace.SpanExporter
 		otlptracegrpc.WithEndpoint(config.Url),
 		otlptracegrpc.WithHeaders(headers),
 	}
-	if config.Tls.Enabled {
-		if tlsCfg != nil {
-			opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
-		}
-	} else {
+
+	if !config.Tls.Enabled {
 		opts = append(opts, otlptracegrpc.WithInsecure())
+		return otlptracegrpc.New(ctx, opts...)
+	}
+	if tlsCfg != nil {
+		opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
 	}
 
 	return otlptracegrpc.New(ctx, opts...)
--- a/pkg/sparrow/metrics/exporter.go
+++ b/pkg/sparrow/metrics/exporter.go
@@ -100,8 +100,10 @@ func newHTTPExporter(ctx context.Context, config *Config) (sdktrace.SpanExporter
 		otlptracehttp.WithEndpoint(config.Url),
 		otlptracehttp.WithHeaders(headers),
 	}
-	if tlsCfg != nil {
-		opts = append(opts, otlptracehttp.WithTLSClientConfig(tlsCfg))
+	if config.Tls.Enabled {
+		if tlsCfg != nil {
+			opts = append(opts, otlptracehttp.WithTLSClientConfig(tlsCfg))
+		}
 	} else {
 		opts = append(opts, otlptracehttp.WithInsecure())
 	}
@@ -120,8 +122,10 @@ func newGRPCExporter(ctx context.Context, config *Config) (sdktrace.SpanExporter
 		otlptracegrpc.WithEndpoint(config.Url),
 		otlptracegrpc.WithHeaders(headers),
 	}
-	if tlsCfg != nil {
-		opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
+	if config.Tls.Enabled {
+		if tlsCfg != nil {
+			opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
+		}
 	} else {
 		opts = append(opts, otlptracegrpc.WithInsecure())
 	}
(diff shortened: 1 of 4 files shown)


y-eight (approved):

pkg/sparrow/metrics/exporter.go
```diff
@@ -120,8 +122,10 @@ func newGRPCExporter(ctx context.Context, config *Config) (sdktrace.SpanExporter
 		otlptracegrpc.WithEndpoint(config.Url),
 		otlptracegrpc.WithHeaders(headers),
 	}
-	if tlsCfg != nil {
-		opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
+	if config.Tls.Enabled {
+		if tlsCfg != nil {
```
puffitos, line 126: Please merge into `if config.Tls.Enabled && tlsCfg != nil`

pkg/sparrow/metrics/exporter.go
```diff
@@ -100,8 +100,10 @@ func newHTTPExporter(ctx context.Context, config *Config) (sdktrace.SpanExporter
 		otlptracehttp.WithEndpoint(config.Url),
 		otlptracehttp.WithHeaders(headers),
 	}
-	if tlsCfg != nil {
-		opts = append(opts, otlptracehttp.WithTLSClientConfig(tlsCfg))
+	if config.Tls.Enabled {
```
puffitos, line 103: Please merge into `if config.Tls.Enabled && tlsCfg != nil`. We can avoid a nested if that way, since there actually is no real need for it, without sacrificing readability.

puffitos (changes requested):
I'll take a better look later, for now everything seems A-OK 👍 Maybe we should introduce a struct for the commonConfig between the exporters (in `getCommonConfig`); returning two totally different instances seems like a code smell we can avoid easily and gain some flexibility by doing so. For a later PR.

lvlcn-t (approved):
When you've implemented @puffitos suggestions it LGTM.

pkg/sparrow/metrics/exporter.go
niklastreml: This would add the insecure option even if `config.Tls.Enabled` is true. I can invert the condition to this, if you want to reduce nesting:
```go
	if !config.Tls.Enabled {
		opts = append(opts, otlptracegrpc.WithInsecure())
		return otlptracegrpc.New(ctx, opts...)
	}
	if tlsCfg != nil {
		opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
	}

	return otlptracegrpc.New(ctx, opts...)
```
I don't think this is necessarily better or worse for readability, but checking boolean flags for falsiness feels weird to me

pkg/sparrow/metrics/exporter.go
niklastreml: Same as [here](https://github.com/caas-team/sparrow/pull/212#discussion_r1804336193)

pkg/sparrow/metrics/exporter.go
puffitos: I didn't mean to remove the else branch, we can definitely keep it. It's one of those rare use cases of the else branch, IMO.

```go
	if config.Tls.Enabled && tlsCfg != nil {
		opts = append(opts, otlptracehttp.WithTLSClientConfig(tlsCfg))
	} else {
		opts = append(opts, otlptracehttp.WithInsecure())
	}
```

Cases:
- If tls is enabled and tlsCfg -> TLSConfig
- if tls is enabled and no tlsCfg -> insecure
- if tls is disabled and tlsCfg -> insecure
- if tls is disabled and no tlsCfg -> insecure

Your suggestion is also fine by me, looks more golangy TBH. Since there are not so many return statements in this functions, it's also easy to understand, even though it introduces implicit logic (if we don't return early, it means tls IS enabled).

Check for falsiness does feel wird sometimes, yes :) Feel free to implement it any way you like, or your way, if I messed up the logic table.

niklastreml: The issue is, that we only want the insecure option when tls is explicitly disabled. Only if the tlsConfig is specified and tls is enabled do we want to load a custom cert chain. The default behaviour of the otel client is to use TLS with the system certs. The table we want looks like this:
| Enabled | *TLSConfig   | Result                              |
|---------|--------------|-------------------------------------|
| True    | &TLSConfig{} | WithTLSClientConfig()               |
| True    | nil          | Do nothing/Use Default Certificates |
| False   | &TLSConfig{} | WithInsecure()                      |
| False   | nil          | WithInsecure()                      |

puffitos: With that information I definetely think your option is the way to go. Since otel offers the New functions and the opts slice, and one option of the opts slice is incompatible with another, an early return on an insecure case configuration is a valid option.

puffitos: I meant as your way the one you proposed:

```go
	if !config.Tls.Enabled {
		opts = append(opts, otlptracegrpc.WithInsecure())
		return otlptracegrpc.New(ctx, opts...)
	}
	if tlsCfg != nil {
		opts = append(opts, otlptracegrpc.WithTLSCredentials(credentials.NewTLS(tlsCfg)))
	}

	return otlptracegrpc.New(ctx, opts...)
```

@niklastreml

puffitos (approved):

Issue #230, opened 2024-12-09
Bug: Gitlab Target Manager does not paginate results in `fetchFileList`


niklastreml:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When fetching the registered targets from gitlab, we never paginate the api. This means we're just taking the default `page_size` of 20 and ignore all other targets that might exist.

### Expected Behavior

Sparrow should find all targets, regardless of how many there might be. 

This shouldn't be a crazy difficult thing to fix. We just need to add a few [query parameters](https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree:~:text=with%20keyset%20pagination.-,pagination,-string) to the url, and look at the `Link` header in the response. If we get the header, just call the url in the header to get the next page. Do this until the header is not returned again and we're happy. This is all described in the [gitlab documentation](https://docs.gitlab.com/ee/api/rest/index.html#keyset-based-pagination). 

### Steps To Reproduce

1. Create a gitlab repo with more than 20 targets
2. Configure the target manager to scan that repo
3. Observe how only some (20 to be exact) of the targets are actually found
4. Profit?

### Who can address the issue?

Any Dev

### Anything else?

Insert fix [here](https://github.com/caas-team/sparrow/blob/main/pkg/sparrow/targets/remote/gitlab/gitlab.go#L144)

https://docs.gitlab.com/ee/api/rest/index.html#keyset-based-pagination
https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree:~:text=with%20keyset%20pagination.-,pagination,-string

lvlcn-t:
Seems to already be fixed with #232.

PR #232, opened 2024-12-11
Feat/tm gitlab  pagination

y-eight:
## Motivation

Referes to #230 

This PR adds the ability to the GitLab targetManager to fetch files recursively in case the GitLab API provides a paginated response.

## Changes

- requests for result pages are made recursively

Additional changes:
- Add helper function to get the next page link of pagination
- Increased the result per page to 30 items (default by GitLab API is 20)
- The `ref` is now added by the `query.Set()` instead of `query.Add()` -> result: query parameter is switched and not appended

## Tests done

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO

- [x] adding test case to test recursive `fetchNextFileList` func

### DIFF
--- a/internal/helper/pagination.go
+++ b/internal/helper/pagination.go
@@ -1,0 +1,52 @@
+// sparrow
+// (C) 2024, Deutsche Telekom IT GmbH
+//
+// Deutsche Telekom IT GmbH and all other contributors /
+// copyright owners license this file to you under the Apache
+// License, Version 2.0 (the "License"); you may not use this
+// file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package helper
+
+import (
+	"net/http"
+	"strings"
+)
+
+const (
+	linkHeader = "Link"
+	linkNext   = "next"
+)
+
+// getNextLink returns the url to the next page of
+// a paginated http response provided in the passed response header.
+func GetNextLink(header http.Header) string {
+	link := header.Get("Link")
+	if link == "" {
+		return ""
+	}
+
+	for _, link := range strings.Split(link, ",") {
+		linkParts := strings.Split(link, ";")
+		if len(linkParts) != 2 {
+			continue
+		}
+		linkType := strings.Trim(strings.Split(linkParts[1], "=")[1], "\"")
+
+		if linkType != linkNext {
+			continue
+		}
+		return strings.Trim(linkParts[0], "< >")
+	}
+	return ""
+}
--- a/pkg/sparrow/targets/remote/gitlab/gitlab.go
+++ b/pkg/sparrow/targets/remote/gitlab/gitlab.go
@@ -177,6 +177,8 @@ func (c *client) fetchNextFileList(ctx context.Context, reqUrl string) ([]string
 		log.ErrorContext(ctx, "Failed to create request", "error", err)
 		return nil, err
 	}
+	req.Header.Add("PRIVATE-TOKEN", c.config.Token)
+	req.Header.Add("Content-Type", "application/json")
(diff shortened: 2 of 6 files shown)


internal/helper/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		linkType := strings.Trim(strings.Split(linkParts[1], "=")[1], "\"")
+
+		if linkType != linkNext {
+			continue
+		}
+		return strings.Trim(linkParts[0], "< >")
+	}
+	return ""
+}
```
lvlcn-t, line 52: I think you can move this to the gitlab package since we don't need it anywhere else. And if we do need it for something else we can move this back to the helper package.

pkg/sparrow/targets/remote/gitlab/gitlab.go
```diff
@@ -26,15 +26,19 @@ import (
...
 	"time"
 
 	"github.com/caas-team/sparrow/pkg/checks"
 	"github.com/caas-team/sparrow/pkg/sparrow/targets/remote"
 
+	"github.com/caas-team/sparrow/internal/helper"
 	"github.com/caas-team/sparrow/internal/logger"
 )
 
+const paginationPerPage = 30 // The amount of items the paginated request to gitlab should return
```
lvlcn-t, line 40: Format the comment as godoc of the constant.
```go
// paginationPerPage is the amount of items the paginated request to gitlab should return
const paginationPerPage = 30
```

internal/helper/pagination.go
y-eight: Oke, moved to gitlab package (y)

y-eight:
Test for paginated requests added.

PR is complete from my side.

pkg/sparrow/targets/remote/gitlab/gitlab.go
```diff
@@ -143,27 +147,37 @@ func (c *client) fetchFile(ctx context.Context, f string) (checks.GlobalTarget,
...
+	log.DebugContext(ctx, "Preparing to fetch file list from gitlab")
+
+	rawUrl := fmt.Sprintf("%s/api/v4/projects/%d/repository/tree", c.config.BaseURL, c.config.ProjectID)
+	reqUrl, err := url.Parse(rawUrl)
+	if err != nil {
+		log.ErrorContext(ctx, "Could not parse GitLab API repository URL", "url", rawUrl, "error", err)
+	}
+	query := reqUrl.Query()
+	query.Set("ref", c.config.Branch)
+	query.Set("per_page", strconv.Itoa(paginationPerPage))
```
niklastreml, line 159: I think we should use [keyset pagination](https://gitlab.devops.telekom.de/caas/microservices/orderportal-2-0/-/issues/2) here instead of offset-based pagination. According to gitlab it's more efficient. Shouldn't be a huge change I think. I think you'd only have to add another query param to it for the initial request, after we can just continue using the Link header.

pkg/sparrow/targets/remote/gitlab/pagination.go
```diff
@@ -0,0 +1,52 @@
...
+
+package gitlab
+
+import (
+	"net/http"
+	"strings"
+)
+
+const (
+	linkHeader = "Link"
```
puffitos, line 27: unused

pkg/sparrow/targets/remote/gitlab/gitlab.go
puffitos: https://docs.gitlab.co.jp/ee/api/rest/index.html#keyset-based-pagination

Yeah, and we should add the `order_by=id&sort=asc` to have some kind of determinist fetching.

puffitos: Additionally, the gitlab token header isn't being set anymore. Am I missing something?

y-eight: Switched to [keyset pagination](https://docs.gitlab.com/ee/api/rest/index.html#keyset-based-pagination) & added sorting parameters.

y-eight: Nice catch @puffitos. Added auth header again & added test coverage

puffitos (approved):
LGTM. I ran the sparrow with `paginationPerPage` set to 20 and still got 30 targets.

Issue #231, opened 2024-12-09
Bug: File conflicts are not handled correctly


niklastreml:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If a registration file remains in a gitlab repo, like after an unsuccessful shutdown, sparrow fails to re-register at the next boot. This happens, because we're POSTing the new file against the API, which gitlab views as a conflict, as it already exists. Gitlab returns a 400 error, and sparrow fails the initial registration.

### Workaround
Until this is fixed, one can manually remove the old registration file from the repo, which should allow the sparrow instance to register again. 

### Expected Behavior

1. For now we should print a proper log output that "a file with this name already exits". 

Gitlab API returns a 400 with an error:

```json
{
  "message": "A file with this name already exists"
} 
```

2. A metric should show the registration status of the sparrow (registered or not). In case the registration failed the metric should be false/0. 

Edit: We decided not to overwrite the file that is already present. This should also prevent from choosing the same name for different sparrows twice. We decided that the failing registration is the default behaviour for now.

### Steps To Reproduce

1. Create a test repo
3. Commit a registration file for your sparrow
```json
{"url":"http://my-sparrow.com","lastSeen":"2024-12-09T15:52:51.763660555Z"}
```
4. Start a sparrow which would try to overwrite this file:
```yaml
name: sparrow.caas-t02.telekom.de
targetManager:
  checkInterval: 1m
  enabled: true
  gitlab:
    baseUrl: https://gitlab.com
    projectId: 123456
  registrationInterval: 10m
  scheme: https
  type: gitlab
  unhealthyThreshold: 48h
  updateInterval: 5h
```

5. Observe logs

### Relevant logs and/or screenshots, environment information, etc.

```console
{"time":"2024-12-09T15:04:37.426700882Z","level":"ERROR","source":{"function":"github.com/caas-team/sparrow/pkg/sparrow/targets.(*manager).register","file":"/home/runner/work/sparrow/sparrow/pkg/sparrow/targets/manager.go","line":180},"msg":"Failed to register global gitlabTargetManager","error":"request failed, status is 400 Bad Request"}
[...]

PR #246, opened 2024-12-25
Feat/decode gitlab error

puffitos:
## Motivation

Relates to #231 

## Changes

Added a general error struct, in which we can decode API error messages and the non-OK status, to return as an error. Additionally, I've added contexts to the target manager's logging calls.

## Tests done

Recreated a case which can create this POST error, by creating a file which has the same DNS name as my dev sparrow b
but a different URL scheme (http vs https). Getting the following logs now (yes, I know we have duplicate logs, but this shouldn't be addressed in this PR):

```shell
time=19:26:03 level=ERROR source=/Users/puffito/dev/repos/sparrow/pkg/sparrow/targets/remote/gitlab/gitlab.go:321 msg="Failed to post file" status="400 Bad Request"
time=19:26:03 level=ERROR source=/Users/puffito/dev/repos/sparrow/pkg/sparrow/targets/manager.go:207 msg="Failed to register global gitlabTargetManager" error="gitlab API sent an unexpected status code (400) with the following error message: {\"message\":\"A file with this name already exists\"}"
time=19:26:03 level=WARN source=/Users/puffito/dev/repos/sparrow/pkg/sparrow/targets/manager.go:127 msg="Failed to register self as global target" error="gitlab API sent an unexpected status code (400) with the following error message: {\"message\":\"A file with this name already exists\"}"
```

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO

### DIFF
--- a/pkg/sparrow/targets/manager.go
+++ b/pkg/sparrow/targets/manager.go
@@ -106,31 +106,31 @@ func (t *manager) Reconcile(ctx context.Context) error {
 	defer registrationTimer.Stop()
 	defer updateTimer.Stop()
 
-	log.Info("Starting target manager reconciliation")
+	log.InfoContext(ctx, "Starting target manager reconciliation")
 	for {
 		select {
 		case <-ctx.Done():
-			log.Error("Error while reconciling targets", "err", ctx.Err())
+			log.ErrorContext(ctx, "Error while reconciling targets", "error", ctx.Err())
 			return ctx.Err()
 		case <-t.done:
-			log.Info("Target manager reconciliation stopped")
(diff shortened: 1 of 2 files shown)


niklastreml (approved):
LGTM, just wondering why you're not exposing the apiError type and chose to keep it private

lvlcn-t (approved):

puffitos:
> LGTM, just wondering why you're not exposing the apiError type and chose to keep it private

Not needed anywhere else and it's extremely specific to the gitlab specification. Looks like a pretty solid candidate for an unexposed type. 😊

PR #266, opened 2025-02-14
fix: camel case for dns result

Swaggeroo:
## Motivation

The DNS check was the only one not fully using camel case in the API response.

## Changes

I added JSON tags to the DNS result struct.

## Tests done

- [x] Unit tests succeeded
- [x] E2E tests succeeded

## TODO

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+<!--
+SPDX-FileCopyrightText: 2025 Deutsche Telekom IT GmbH
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
 # Changelog
 
 ## [0.6.0](TODOADDRELEASE) (DATE)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## [0.6.0](TODOADDRELEASE) (DATE)
+
+### Features
+
+* added Changelog file
+
+### Bug Fixes
+
+* **BREAKING CHANGE:** DNS metric returns lowercase keys instead of uppercase ones
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,10 @@ SPDX-License-Identifier: CC-BY-4.0
 
 # Changelog
 
-## [0.6.0](TODOADDRELEASE) (DATE)
+## [v0.6.0](TODOADDRELEASE) (DATE)
+⚠️ This release contains potential breaking change ⚠️
 
-### Features
+The API returns lowercase keys instead of capitalized keys. Ensure that your code handles this change to avoid issues.
 
-* added Changelog file
-
-### Bug Fixes
-
-* **BREAKING CHANGE:** DNS metric returns lowercase keys instead of uppercase ones
+* [ENHANCEMENT] added Changelog file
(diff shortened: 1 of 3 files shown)


pkg/checks/dns/dns.go
```diff
@@ -60,9 +60,9 @@ func NewCheck() checks.Check {
 
 // result represents the result of a single DNS check for a specific target
 type result struct {
-	Resolved []string
-	Error    *string
-	Total    float64
+	Resolved []string `json:"resolved"`
+	Error    *string  `json:"error"`
+	Total    float64  `json:"total"`
```
lvlcn-t, line 65: Although I'm not against this change (since sparrow is still considered unstable), we should document it as a breaking change. Please add a note in the README to inform users who may rely on the previous API response format. Furthermore a note in the next release notes should also be added then.

lvlcn-t (changes requested):

CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
+<!--
+SPDX-FileCopyrightText: 2025 Deutsche Telekom IT GmbH
+
+SPDX-License-Identifier: CC-BY-4.0
+-->
+
+# Changelog
+
+## [0.6.0](TODOADDRELEASE) (DATE)
```
lvlcn-t, line 9: Add the `v` prefix here.

CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Features
+
+* added Changelog file
+
+### Bug Fixes
+
+* **BREAKING CHANGE:** DNS metric returns lowercase keys instead of uppercase ones
```
lvlcn-t, line 17: Can you mention this immediately at the beginning for this release? The user should be able to directly see it.

You may want to take a look at the last [prometheus go client](https://github.com/prometheus/client_golang/blob/main/CHANGELOG.md#1210--2025-02-17) changelog to get some inspiration.

README.md
```diff
@@ -465,6 +465,8 @@ latency:
   - Labelled with `target`
 
 ### Check: DNS
+> [!CAUTION]
+> Versions > v0.5.0, the api returns lowercase keys instead of uppercase.
```
lvlcn-t, line 469: Here are some problems: 
1. `v0.6.0` instead of the already released `v0.5.0`.
2. I'd mention the words _breaking change_ to better highlight that the user needs to do something if he uses the API.
3. The old keys were not uppercase but rather capitalized.

For example you could write something like this:
```suggestion
### Check: DNS
> [!CAUTION]
> **Breaking Change:** Starting from version `v0.6.0`, the API returns lowercase keys instead of capitalized keys. Ensure that your code handles this change to avoid issues.
```

README.md
```diff
@@ -630,6 +632,9 @@ Is roughly equal to this:
 
 ## API
 
+> [!CAUTION]
+> Versions > v0.5.0, the api returns lowercase keys instead of uppercase for DNS metric.
+
```
lvlcn-t, line 637: Here again the same problems as above.

```suggestion
> [!CAUTION]
> **Breaking Change:** Starting from version `v0.6.0`, the API for the DNS check returns lowercase keys instead of capitalized keys. Ensure that your code handles this change to avoid issues.
```

lvlcn-t (changes requested):

y-eight (approved):

PR #278, opened 2025-03-23
refactor(checks): move global target merging responsibility to checks

### LINKED ISSUE #100: Refactor: Move global target merging responsibility to checks

### Problem Description

The global targets are injected to the checks config into their `target` field. This was ok for the target manager MVP but the checks will become more complex and not every check can implement a string slice as target field.

### Solution Description

Add a dedicated global targets field to the checks config structs. This field can then be merged into the targets field by the check itself depending on its config implementation.

Everyone

### Additional Context

_No response_

lvlcn-t:
## Motivation

Closes #100 

I've also added global target merging for the traceroute check, as we didn't enrich the targets yet.

## Changes

* [refactor(checks): move global target merging responsibility to checks](https://github.com/telekom/sparrow/commit/a3e6d4a6955bc8b0ff14e72e32e4da3839bb1e7b) 

This commit moves the responsibility of merging global targets with the configured targets to each check. This way the checks can decide how to merge the global targets with the configured targets.

Signed-off-by: lvlcn-t <75443136+lvlcn-t@users.noreply.github.com>

## Tests done

I've added an additional test case for the traceroute enrichment.

- [x] Unit tests succeeded
- [ ] ~E2E tests succeeded~ - we have none for this

## TODO

### DIFF
--- a/pkg/checks/base.go
+++ b/pkg/checks/base.go
@@ -6,6 +6,8 @@ package checks
 
 import (
 	"context"
+	"net/url"
+	"strconv"
 	"sync"
 	"time"
 
@@ -56,12 +58,15 @@ type CheckBase struct {
 	DoneChan chan struct{}
 }
 
-// Runtime is the interface that all check configurations must implement
+// Runtime is the interface that all check configurations must implement.
 type Runtime interface {
-	// For returns the name of the check being configured
+	// For returns the name of the check being configured.
 	For() string
-	// Validate checks if the configuration is valid
+	// Validate checks if the configuration is valid.
 	Validate() error
+	// Enrich enriches the configuration with the [GlobalTarget]s.
+	// Each configuration is responsible for adding the targets it needs.
+	Enrich(ctx context.Context, targets []GlobalTarget)
 }
 
 // Result encapsulates the outcome of a check run.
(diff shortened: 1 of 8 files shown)


pkg/checks/base.go
```diff
@@ -84,3 +89,50 @@ type GlobalTarget struct {
 	Url      string    `json:"url"`
 	LastSeen time.Time `json:"lastSeen"`
 }
+
+// URL returns the [url.URL] representation of the target.
+func (g GlobalTarget) URL() (*url.URL, error) {
+	return url.Parse(g.Url)
+}
```
niklastreml, line 96: We're already storing the url on the GlobalTarget as a string. Why not parse that string into a url when we first construct the object? That way getting the url can never fail. Getting the URL as a string is using stringbuilder internally, so it should also be reasonably efficient (atleast faster than parsing the url)

niklastreml: To add to this, we also call this method everywhere in the interface implementation, so parsing the url is done quite often.

lvlcn-t: This will require a bigger refactoring, since we use the string **everywhere**. Should I do it in this PR or shall I open another?

PR #433, opened 2026-05-10
feat: add polling jitter to target manager intervals

lvlcn-t:
## Motivation

When many sparrow instances share the same polling intervals, they
synchronize into a thundering herd — all hitting the backend at the same
instant. Adding jitter spreads requests over time, reducing peak load on
the S3/GitLab backend and improving resilience.

**Stack:** PR 4/4 — builds on #432 (OIDC auth).

## Changes

- Add `helper.ApplyJitter(d, factor)` in `internal/helper/jitter.go` with
  bounded minimum `[d*(1-factor), d]`
- Guard against `NaN` and `Inf` factor inputs (treat as identity)
- Add `Jitter float64` field to `General` config struct with `[0.0, 1.0]`
  validation and `ErrInvalidJitter` sentinel
- Wire jitter into `manager.Reconcile()` on all three timer resets
  (check, registration, update)
- Add `FuzzApplyJitter` fuzz test (~3.7M executions, no failures) with
  property-based assertions (bounded minimum, never exceeds original,
  non-negative result)
- Add deterministic `TestApplyJitter_EdgeCases` for NaN/Inf/zero
- Add 4 validation test cases for jitter bounds
- Document jitter in `chart/values.yaml`

## Tests done

- [x] Unit tests succeeded
- [x] Fuzz tests succeeded (~3.7M executions)
- [ ] E2E tests succeeded

## TODO

- [ ] I've assigned this PR to myself
- [ ] I've labeled this PR correctly

### DIFF
--- a/internal/helper/jitter.go
+++ b/internal/helper/jitter.go
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText: 2025 Deutsche Telekom IT GmbH
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package helper
+
+import (
+	"math/rand/v2"
+	"time"
+)
+
+// ApplyJitter applies full jitter with a bounded minimum to a duration.
+// The returned duration is in the range [d*(1-factor), d].
+// A factor of 0 returns d unchanged. Factor must be in [0.0, 1.0].
+func ApplyJitter(d time.Duration, factor float64) time.Duration {
+	if factor <= 0 || d <= 0 {
+		return d
+	}
+	if factor > 1 {
+		factor = 1
+	}
+
+	minDuration := float64(d) * (1 - factor)
+	jitterRange := float64(d) * factor
+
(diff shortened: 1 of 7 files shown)


internal/helper/jitter.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"math"
+	"math/rand/v2"
+	"time"
+)
+
+// ApplyJitter applies full jitter with a bounded minimum to a duration.
+// The returned duration is in the range [d*(1-factor), d].
+// A factor of 0 returns d unchanged. Factor must be in [0.0, 1.0].
```
niklastreml: this docstring isnt helpful for the caller. leave it as a function comment.

For the caller maybe say something like "d is the duration, factor is a percentage of duration to add to the duration.

Also consider if we want this to also apply a negative duration so the the call could be up to factor% faster than `d`. Not sure if we want this or not

lvlcn-t: I've kept the positive only duration for now, we can change it if we have to later on.
Fixed the comment.

